PR #11, opened 2015-11-05
Add file logger to capture activity

abought:
Description: https://openscience.atlassian.net/browse/OSF-5085

## Purpose
(relevant) User activity should be logged to a file, not just console- this will aid in providing support and capturing incidents

## Summary of changes
- Add a file handler to log all activity to a single file (initial proof of concept: ~/Library/Logs/osf-offline/osfoffline.log)
- Reduce verbosity of logs: general flow events reduced to debug so they won't appear in normal user activity logs
- Move centralized settings/setup to a module
- Added custom sys.excepthook to capture uncaught exceptions as part of log (needs to be tested)
- Added / enhanced testing options in tasks.py `start_for_tests`

## To do
- [x] Review to ensure we aren't logging passwords or other sensitive data to file


osfoffline/settings/app_settings.py
```diff
@@ -0,0 +1,33 @@
...
+    FILE_BASE = 'http://localhost:7777'  ## FIXME: Dev mode currently does not work with local waterbutler (abought)
+else:
+    API_BASE = 'https://staging-api.osf.io'
+    FILE_BASE = 'https://staging-files.osf.io'
+
+### Interval (in seconds) to poll the OSF for server-side file changes
+if DEV_MODE is True:
+    POLL_DELAY = 5  # seconds
+else:
+    POLL_DELAY = 15 * 60  # seconds
```
chennan47: Aim the delay to be 5 mins, will make this change when I merge.

osfoffline/application/main.py
```diff
@@ -138,7 +136,7 @@ def start(self):
         self.background_worker.start()
 
     def resume(self):
-        logging.info("resuming")
+        logging.debug("resuming")
```
chennan47: single quote, will make this change when i merge.

tasks.py
```diff
@@ -53,13 +54,23 @@ def start():
...
 
 @task
-def start_for_tests():
-    if os.path.exists(DB_FILE_PATH):
-        os.remove(DB_FILE_PATH)
-
-    OSF_DIR = '~/Desktop/OSF'
-    if os.path.exists(OSF_DIR):
-        shutil.rmtree(OSF_DIR)
+def start_for_tests(dropdb=True, droplog=False, dropdir=False):
```
chennan47: add option to change the polling_delay timer and add logging options since the dev_mode is not quite working with local copy of osf. will make the change when I merge.

PR #13, opened 2015-11-05
[WIP][Feature] Authorization via Personal Access Tokens

mfraezz:
Purpose
=======
Provide a simpler way of authorizing than using cookies

Changes
=======
* Start screen UI updated to accept only PAT, with instructions on PAT generation
* Login method updated to use PATs

Side Effects
=========
* better `[(<local>, <remote>), ...]` tuple list creation logic
* better stop_loop logic
* more logging added

Notes
=====
* Tested to be functional on Mac and Windows
* TODO: better error handling related to login timeouts

[#OSF-5064]
[#OSF-5096]


osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,65 @@ def __init__(self):
...
-            logging.warning('user doesnt exist. Creating user. and logging them in.')
-            user = User(
-                full_name=full_name,
-                osf_id=osf_id,
-                osf_login=user_name,
-                osf_local_folder_path='',
-                oauth_token=oauth_token,
-                osf_password=password
+            resp = yield from osf.make_request(url)
+        except (aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError, aiohttp.errors.HttpBadRequest, concurrent.futures._base.TimeoutError):
```
abought: Split these into separate exceptions with different error messages depending on the nature of the problem

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,65 @@ def __init__(self):
...
+
+            else:
+                # Authentication failed, user has likely provided an invalid token
+                raise OSFAuthError('Invalid auth response: {}'.format(resp.status))
 
             self.close()
 
     def setup_slots(self):
-        self.start_screen.logInButton.clicked.connect(self.log_in)
+        logging.info('setting up start_screen slots')
```
abought: This is an internal behavior that isn't necessary for the user to see/ know about. Change the log level to "debug".

osfoffline/application/background.py
```diff
@@ -62,31 +62,30 @@ def get_current_user(self):
...
-
-        elif not self.loop.is_running():
-            logging.info('loop is stopped already. closing it')
-            self.loop.close()
-        else:
-            # stop loop when current tasks finish.
-            self.loop.call_soon(self.loop.stop)
-            logging.info('call_soon to loop.stop. will stop when polling/observing events finish.')
-            # todo: find better way?
+            logging.info('loop already stopped')
```
abought: This can be changed to either `logging.debug` or removed entirely as appropriate

abought: Dittio for other logging events throughout this PR; I'll try to avoid spamming with a comment on each.

osfoffline/application/background.py
```diff
@@ -62,31 +62,30 @@ def get_current_user(self):
...
             if close:
-                while not self.loop.is_closed():
-                    if not self.loop.is_running():
-                        self.loop.close()
+                logging.info('closing loop')
```
abought: This particular log statement may be unnecessary (log that it's closing if loop is closed)

mfraezz: Stopped isn't the same as closed. There's supposed to be a call to `self.loop.close()` below here, adding that.

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -19,8 +19,8 @@
 class OSFQuery(object):
     def __init__(self, loop, oauth_token):
         self.headers = {
-            # 'Authorization': 'Bearer {}'.format(oauth_token),
-            'Cookie': 'osf_staging={}'.format(oauth_token)
+            'Authorization': 'Bearer {}'.format(oauth_token),
+            # 'Cookie':'osf_staging={}'.format(oauth_token)
```
abought: Remove the line referencing cookie. If we want it back we can look at the git history.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,72 @@ def __init__(self):
...
-                osf_password=password
+            resp = yield from osf.make_request(url)
+        except (aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError, concurrent.futures._base.TimeoutError):
+            # No internet connection
+            QMessageBox.warning(
+                None,
+                "Log in Failed",
+                "Unable to connect to server. Check your internet connection or try again later."
+            )
+            self.hide()
```
abought: Hiding and reopening the window seems unnecessary. (and leads to a subtle flash on screen)

osfoffline/application/background.py
```diff
@@ -19,6 +19,7 @@ def __init__(self):
         self.paused = True  # start out paused
         self.running = False
         self.poller = None
+        self.observer = None
```
abought: Where in this PR is `self.observer` set to something other than None?

abought: Ah, I see: it was always going to be set later; this just makes sure we include this attribute as part of the definition of the class.

Very helpful- thanks! :octopus:

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -278,8 +277,9 @@ def make_request(self, url, method=None, params=None, expects=None, get_json=Fal
                 raise aiohttp.errors.BadStatusLine(response.status)
         elif 400 <= response.status < 600:
             content = yield from response.read()
-            error_message = '[status code: {}]:: {} @url '.format(str(response.status), str(content), str(url))
+            error_message = '[status code: {}]:: {} @url {}'.format(str(response.status), str(content), str(url))
             logging.error(error_message)
+            self.close()
             raise aiohttp.errors.HttpBadRequest(error_message)
```
abought: Ah- here is the place where the authorization error gets masked into a generic "Bad Request" exception.
(per our discussion of https://github.com/himanshuo/OSF-Offline/pull/13/files#diff-1d5902bdc5bcd3fa27c12b42ae6ff926R51)

Longterm this should be improved as it may mask edge cases. For now we'll leave it to ensure we can ship an MVP.

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -278,8 +277,9 @@ def make_request(self, url, method=None, params=None, expects=None, get_json=Fal
                 raise aiohttp.errors.BadStatusLine(response.status)
         elif 400 <= response.status < 600:
             content = yield from response.read()
-            error_message = '[status code: {}]:: {} @url '.format(str(response.status), str(content), str(url))
+            error_message = '[status code: {}]:: {} @url {}'.format(str(response.status), str(content), str(url))
```
abought: `.format` will handle type coercion; no need for the `str` calls.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
-            logging.warning('user doesnt exist. Creating user. and logging them in.')
-            user = User(
-                full_name=full_name,
-                osf_id=osf_id,
-                osf_login=user_name,
-                osf_local_folder_path='',
-                oauth_token=oauth_token,
-                osf_password=password
+            resp = yield from osf.make_request(url)
+        except (aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError, concurrent.futures._base.TimeoutError):
```
abought: You can just import `TimeOutError` from `aiohttp.errors` directly (it's aliased to the same thing)

osfoffline/polling_osf_manager/polling.py
```diff
@@ -110,48 +110,21 @@ def make_local_remote_tuple_list(self, local_list, remote_list):
         assert None not in local_list
         assert None not in remote_list
 
-        combined_list = local_list + remote_list
-        sorted_combined_list = sorted(combined_list, key=self.get_id)
+        local_files = {}
```
abought: Himanshu has approved this rewrite, but we'll be sure to test the jeepers out of it. :shipit:

abought:
For a followon ticket, the personal access token input box should be treated as a password field. We want users to feel that we are taking security precautions.

osfoffline/views/preferences.py
```diff
@@ -147,8 +147,7 @@ def get_remote_top_level_nodes(self):
             if user:
                 user_nodes = []
                 url = api_url_for(USERS, related_type=NODES, user_id=user.osf_id)
-                # headers={'Authorization': 'Bearer {}'.format(user.oauth_token)}
-                headers = {'Cookie': 'osf_staging={}'.format(user.oauth_token)}
+                headers = {'Authorization': 'Bearer {}'.format(user.oauth_token)}
                 resp = requests.get(url, headers=headers).json()
                 logging.warning(resp)
```
abought: Followon comment: we should not need need to log the entire JSON payload to the console, much less as a warning.

osfoffline/views/start_screen.py
```diff
@@ -1,12 +1,22 @@
+import furl
+import aiohttp
```
abought: Per PEP8, move installed packages below standard library modules. (furl and aiohttp below asyncio etc)

Within each category of imports, generally try to keep the imports alphabetical, to make it easier to find if/where something is imported.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+                "Invalid login credentials."
             )
-            user.logged_in = True
-            save(session, user)
+        else:
+            if resp.status == 200:
+                json_resp = yield from resp.json()
+                full_name = json_resp['data']['attributes']['full_name']     # Use one of these later to "Welcome {user}"
+                given_name = json_resp['data']['attributes']['given_name']   # Use one of these later to "Welcome {user}"
+                osf_id = json_resp['data']['id']
```
abought: Change this to use the RemoteUser class (which provides a centralized abstraction for parsing the data):
https://github.com/abought/OSF-Offline/blob/develop/osfoffline/polling_osf_manager/remote_objects.py#L25-25

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+
+                try:
+                    user = session.query(User).filter(User.osf_id == osf_id).one()
+                    user.logged_in = True
+                    save(session, user)
```
abought: This try/except block contains way too much stuff. It should only contain the query information sufficient to `find` the user- it should not mutate the user (change login state, save, etc).

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+                except MultipleResultsFound:
+                    logging.warning('multiple users with same username. deleting all users with this username. restarting function.')
```
abought: Is "deleting everything we know about that user" the best way to handle this problem? Wouldn't this cause a data integrity issue?

This is particularly important as we move to personal access tokens- if someone generates a new token, it's very possible to have the same user in the DB twice.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+                        osf_id=osf_id,
+                        osf_login='',  # TODO: email goes here when more auth methods are added, not currently returned by APIv2
+                        osf_local_folder_path='',
+                        oauth_token=personal_access_token,
+                    )
+                    user.logged_in = True
+                    save(session, user)
+
+            else:
+                raise OSFAuthError('Invalid auth response: {}'.format(resp.status))
```
abought: This is an unhandled exception, which means the program will do bad things instead of informing the user that login failed. (compare to the case if it gets a particular bad request error code, here: https://github.com/himanshuo/OSF-Offline/pull/13/files#diff-1d5902bdc5bcd3fa27c12b42ae6ff926R46 )

The UI should not become unavailable if login fails- that

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+                except NoResultFound:
+                    logging.debug('user doesnt exist. Creating user. and logging them in.')
+                    user = User(
+                        full_name=full_name,
+                        osf_id=osf_id,
+                        osf_login='',  # TODO: email goes here when more auth methods are added, not currently returned by APIv2
+                        osf_local_folder_path='',
+                        oauth_token=personal_access_token,
+                    )
+                    user.logged_in = True
```
abought: The `user.logged_in` + `save()` code is duplicated along all these paths regardless of whether a user is found, and should be moved outside the try/except block.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+                    save(session, user)
```
abought: Should we add error handling in the case that `save` (re)raises an exception?

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
+
+            else:
+                raise OSFAuthError('Invalid auth response: {}'.format(resp.status))
 
             self.close()
 
     def setup_slots(self):
-        self.start_screen.logInButton.clicked.connect(self.log_in)
+        logging.debug('setting up start_screen slots')
+        self.start_screen.logInButton.clicked.connect(lambda: asyncio.get_event_loop().run_until_complete(self.log_in()))
```
abought: If I am reading this right, we've written very complicated code that makes an async request (to be non-blocking)... and then converted it into a blocking method?

It seems like there's an extra step there.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +31,68 @@ def __init__(self):
...
-        # assumption: logged in user properly.
-        # assumption: all the fields needed for db are set
-        full_name = user_name
-        osf_id = user_name
-        oauth_token = password
-        # check if user already exists in db
-
+        logging.debug('attempting to log in')
+        personal_access_token = self.start_screen.personalAccessTokenEdit.text().strip()
+        url = settings.API_BASE + '/v2/users/me/'
```
abought: We may have to eat it on this pass... but for future methods, we should not put so much request / parsing/ database logic inside a UI callback- this isn't good separation of concerns, and this method is way more complicated than it probably needs to be.

abought:
CR pass complete. Biggest issues include data handling on error, and a few unraised exceptions.

Also some low-hanging code cleanup by using the shared data parsing class, and simplifying some unwieldy (and potentially buggy) try except blocks.

osfoffline/application/background.py
mfraezz: :+1:

osfoffline/views/start_screen.py
mfraezz: I find triangular import organization more visually appealing (each import line per section is longer than the previous), but can change it to alphabetical if it's very important to you.

osfoffline/views/start_screen.py
mfraezz: Is there a clean way to merge user accounts? If not, we should keep this the way it is and create an improvement ticket.

Also, I'm not certain that even that would create multiple users (as it's filtering by `user.osf_id`, not by `.oauth_token`), but I'm adding logic to update the `oauth_token` if necessary, to handle this case.

osfoffline/views/start_screen.py
mfraezz: Realistically, I don't think this is reachable. `.make_request` is going to raise an error almost all of the time if it's not a `200 OK`. I'll make even more sure of that and remove this `else`.

osfoffline/views/start_screen.py
mfraezz: What exception [c|w]ould it (re)raise? A user is being created here, and unless `session` is broken somehow, it shouldn't raise anything.

osfoffline/views/start_screen.py
mfraezz: When we have a week to make this into a MVP, picking of nits should come after fixing of bugs, at least IMHO.

osfoffline/views/start_screen.py
mfraezz: Need to 'unwrap' the coroutine or it will break everything.

osfoffline/views/start_screen.py
abought: Thanks. Sadly, while visual appeal is fun in personal code, this is about maintainability- triangular imports don't provide the level of predictability needed for code worked on by many people.

As one example, I recently reviewed a module that used `import module`, `from module import x`, and `import module as alias` in the same block. Had imports been ordered, that clutter could have been avoided.

If it makes you feel better, it's the wave of the future; Jetbrains recently made it the [default behavior](http://blog.jetbrains.com/pycharm/2015/09/announcing-the-pycharm-5-eap-opening/) in one of their products.

osfoffline/views/start_screen.py
abought: *nods* Code review isn't always about what we can do today- sometimes it's about ways we can improve for the future. I tried to avoid most nitpick comments and limit myself to a few broader points towards that end.

osfoffline/views/start_screen.py
abought: Post MVP comment:

Yeah... but as far as I can tell, the only reason it needs to be a coroutine is because it depends on `make_request`. 

By using an async method when we didn't need to, we effectively succeeded in making blocking code more complicated than it needs to be (and increased our debugging effort, as you found the hard way).

osfoffline/views/start_screen.py
```diff
@@ -21,47 +32,67 @@ def __init__(self):
...
+            # Invalid credentials
+            QMessageBox.warning(
+                None,
+                "Log in Failed",
+                "Invalid login credentials."
+            )
+        else:
+            json_resp = yield from resp.json()
+            remote_user = RemoteUser(json_resp['data'])
+            full_name = json_resp['data']['attributes']['full_name']     # Use one of these later to "Welcome {user}"
```
abought: This can be changed to `remote_user.name`. If you're going to use an abstracted data parsing class, use it consistently.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +32,67 @@ def __init__(self):
...
+            given_name = json_resp['data']['attributes']['given_name']   # Use one of these later to "Welcome {user}"
+            osf_id = json_resp['data']['id']
```
abought: `remote_user.id`

osfoffline/views/start_screen.py
```diff
@@ -21,47 +32,67 @@ def __init__(self):
...
+            QMessageBox.warning(
+                None,
+                "Log in Failed",
+                "Invalid login credentials."
+            )
+        else:
+            json_resp = yield from resp.json()
+            remote_user = RemoteUser(json_resp['data'])
+            full_name = json_resp['data']['attributes']['full_name']     # Use one of these later to "Welcome {user}"
+            given_name = json_resp['data']['attributes']['given_name']   # Use one of these later to "Welcome {user}"
```
abought: Since there's now a use case for this value, add this to the `RemoteUser` class and eliminate this as a separate variable.

osfoffline/views/start_screen.py
abought: Good point on the improvement ticket. We have some everyday user-facing bugs to triage first, and then we can tackle edge cases that will *create* bugs later.

osfoffline/views/start_screen.py
```diff
@@ -21,47 +32,67 @@ def __init__(self):
...
+                    osf_local_folder_path='',
+                    oauth_token=personal_access_token,
+                )
+                user.logged_in = True
+                save(session, user)
+            else:
+                if not user.oauth_token == personal_access_token:
+                    user.oauth_token == personal_access_token
+                user.logged_in = True
+                save(session, user)
```
abought: The user.logged_in and save calls are still duplicated (see line 83); why not move outside the try-except altogether?

osfoffline/views/start_screen.py
abought: Anything that would force SqlAlchemy to raise an error- data not matching field types, transaction issues, etc. http://docs.sqlalchemy.org/en/latest/core/exceptions.html

If the database failed to save, we should notify the user that the login failed, and they should try again later. There should be a logging statement (logging.error) to indicate that the problem was something internal to the database, rather than a problem with user credentials.

abought:
A couple more exception handling notes/suggestions, and request to use the data parsing class consistently. Passing back, briefly.

(Please don't try to get to this tonight. I'm looking forward to sleeping)

osfoffline/views/start_screen.py
mfraezz: I don't necessarily take the default behaviour of PyCharm as best practice, but I'll do it anyway to appease the CR gods.

mfraezz:
It's here for you in the morning.

osfoffline/views/start_screen.py
abought: Best I can say is that it's a moderately absurd solution to a very absurd situation in our active codebase. I agree that in a perfect world, we could both go back to not caring. For now, at least we'll have company in the insanity. :)

osfoffline/views/start_screen.py
```diff
@@ -21,47 +32,63 @@ def __init__(self):
...
+                    osf_login='',  # TODO: email goes here when more auth methods are added, not currently returned by APIv2
+                    osf_local_folder_path='',
+                    oauth_token=personal_access_token,
+                )
+            else:
+                if not user.oauth_token == personal_access_token:
+                    user.oauth_token == personal_access_token
+
             user.logged_in = True
             save(session, user)
```
abought: The possibility of a database save leaving an exception is the last thing to be addressed. It may have been missed because the conversation built on an outdated diff; see email notification for the UI suggestion.

abought:
The `RemoteUser` object aliases full_name to `'name` (see class definition). Trying to actually run the code gives the exception below:

```
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/views/start_screen.py", line 108, in <lambda>
    self.start_screen.logInButton.clicked.connect(lambda: asyncio.get_event_loop().run_until_complete(self.log_in()))
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/base_events.py", line 342, in run_until_complete
    return future.result()
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(value)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/views/start_screen.py", line 83, in log_in
    full_name=remote_user.full_name,
AttributeError: 'RemoteUser' object has no attribute 'full_name'
```

abought:
Al the Octopus waves you forward to CR2. :octopus: :checkered_flag:

osfoffline/views/rsc/gui/startscreen_gui/startscreen.ui
```diff
@@ -38,50 +38,48 @@
...
+      <width>421</width>
+      <height>111</height>
      </rect>
     </property>
     <property name="text">
-     <string>Password</string>
+     <string>Note: Additional login methods will be available later. 
+
+To generate a personal access token, go to 
+http://osf.io/settings/tokens/create/
```
chennan47: It will be better to make this a clickable link

PR #24, opened 2015-11-07
[hold][Feature] Stop Upload of a Modified File to prevent corruption

mfraezz:
Purpose
======
Handle case where user has pending upload of a file currently being edited, and it changes on disk before upload completes

Changes
=======
Async streaming/locking mechanism added

Side Effects
=========
None

Notes
--------
* WIP tag included because I've been unable to fully test this feature -- have yet to reproduce bug. Have tested to confirm it does not break normal upload flow.
* Consider using `Network Link Conditioner` to test, downloadable through `XCode`->`Open Developer Tool`->`More Developer Tools`-> Search: `Hardware IO Tools`

[#OSF-5107]


osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -242,34 +251,39 @@ def _delete_file_folder(self, remote_file_folder):
...
+            headers=headers,
             params=params,
             data=data
         )
         try:
-            response = yield from asyncio.wait_for(request, timeout)
-        except (
-                aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError,
+            if locks and local_file:
+                asyncio.async(check_for_changes(local_file, locks))
```
abought: As discussed yesterday, `asyncio.async` is deprecated. Use `asyncio.ensure_future` instead.
https://docs.python.org/3/library/asyncio-task.html#asyncio.async

Additionally, it appears that `ensure_future` would return a Task object which can be canceled explicitly- meaning that `check_for_changes` could be canceled/terminated without need for the second lock. Is that correct?

mfraezz: It is correct, the second lock would not be needed. However, as discussed privately, `asyncio.ensure_future` doesn't exist before Python 3.4.4, and this is too much of a headache with current deadlines

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -132,7 +135,13 @@ def upload_file(self, local_file):
...
                                 file_id=parent_osf_id)
         file = open(local_file.path, 'rb')
-        resp_json = yield from self.make_request(files_url, method="PUT", params=params, data=file, get_json=True)
+
+        file_lock = asyncio.Lock()
+        file_data = safe_file_stream(file, file_lock)
+        file_size = os.path.getsize(local_file.path)
+
+        resp_json = yield from self.make_request(files_url, method="PUT", params=params, data=file_data, get_json=True,
+            headers={'Content-Length': str(file_size)}, local_file=local_file, locks=[file_lock, asyncio.Lock()])
```
chennan47: what if there is a http error or server is down during this request. There is no exception being raised here.

chennan47: Oh. My bad, ignore the previous comment.

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -242,34 +251,39 @@ def _delete_file_folder(self, remote_file_folder):
         resp.close()
 
     @asyncio.coroutine
-    def make_request(self, url, method=None, params=None, expects=None, get_json=False, timeout=180, data=None):
+    def make_request(self, url, method=None, params=None, expects=None, get_json=False, timeout=180, data=None, headers=None, local_file=None, locks=None):
```
chennan47: This line is too long. better break  it into multi lines. Will make the change when I merge.

osfoffline/polling_osf_manager/osf_query.py
chennan47: As Matt said, the version of python we are using is python 3.4.3, it won't be too hard to change to python 3.4.4. But we need to set down a version to avoid future's change.

abought:
@brianjgeiger will take point on review. I can help him with some testing notes for this one. Particularly:

- Using the develop branch to edit a file as it is being uploaded, the file isn't corrupted, but neither is it clear that the version sent is the one desired. Will explore further. (regression tests: before AND after)

- Apple provides a tool called Network Link Conditioner that is useful for simulating slow connections. (the "Edge network" preset is handy for testing slow uploads, to give time to see the change occur)

brianjgeiger:
Hey, @mfraezz, can you merge this with the just merged changes from the threading pr?

osfoffline/polling_osf_manager/osf_query.py
mfraezz: Since we've updated to 3.4.4, I'll update this

mfraezz:
@brianjgeiger :+1:

abought:
Is there any update on the progress of review for this ticket, @brianjgeiger ?

brianjgeiger:
I have not yet managed to reproduce the issue. I tried a couple of times before I switched to my task from yesterday of dealing with the error handling.

abought:
Okay. We'll table this one for Tuesday's release then, and revisit later when we have more breathing room.

abought:

PR #25, opened 2015-11-08
Feature/fix items not syncing

brianjgeiger:
Purpose
-----------
Allow a user to drag a file into the root syncing folder, then drag the file into the project folder, and still get the file to sync into the project properly. Fixes [https://openscience.atlassian.net/browse/OSF-5121]

Secondarily, modified a small UI issue: swapped the column order on the preferences tab so that the checkboxes appeared first and the project items appeared second. It was a tiny change but should help immensely until further UI refinements can be added in.

Changes
----------
What was happening is that, once the file was seen by the desktop sync code, moving the file was considered a move, but moving only happens between projects. So I caught the case where it didn't really know where the file was being moved from and just said, well, if you don't know, then it must be a create action instead. I broke up the create into its own method that is separate from the event handler rather than trying to trigger a new event with the existing file.

For the UI change, I changed the ordering numbers in the constant for that control, and then I ensured that the numbers matched for the column labels.

Side effects
---------------
Hopefully none, but ¯\\\_(ツ)_/¯

### DIFF
--- a/osfoffline/filesystem_manager/osf_event_handler.py
+++ b/osfoffline/filesystem_manager/osf_event_handler.py
@@ -5,7 +5,7 @@ storing the data into the db, and then sending a request to the remote server.
 import asyncio
 import logging
 
-from watchdog.events import FileSystemEventHandler, DirModifiedEvent
+from watchdog.events import FileSystemEventHandler, DirModifiedEvent, DirCreatedEvent, FileCreatedEvent
 from osfoffline.database_manager.models import Node, File, User
 from osfoffline.database_manager.db import session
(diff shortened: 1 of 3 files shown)


brianjgeiger:
If you don't have time/energy to deal with the UI thing, feel free to cherry pick the meatier commit.

abought:
Logic works, and verified it works in manual testing. Clear for @chennan47 .

ticklemepierce:
@brianjgeiger I cant figure out that nifty reply feature but yes I am incorporating it into another UI fix.  It is definitely a worthwhile change.

PR #26, opened 2015-11-08
[Feature] Standard Auth

mfraezz:
Purpose
=======
Allow a user to log in with username and password, and use that data to generate a non-expiring personal access token.

Changes
=======
* Add `authentication.py` to handle log in flow
* Add `bcrypt` to `requirements` for password hashing

Side effects
=========
None that I've found ¯\\\_(ツ)_/¯


osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,196 @@
...
+
+    :return: hashed password
+    """
+    return bcrypt.hashpw(
+        password.encode('utf-8'),
+        bcrypt.gensalt(12)
+    )
+
+token_request_body = json.loads(
+    '{"data": {"type": "tokens", "attributes": {"name": "OSF-Offline {}", "scopes": "osf.full_write"}}}'.format(datetime.date.today())
```
abought: Since you're writing the dictionary inline, no need to convert it to a string. Just specify it as a dictionary to start.

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @asyncio.coroutine
+    def _authenticate(self, username, password):
+        """ Tries to use standard auth to authenticate and create a personal access token through APIv2
+
+            :return: personal_access_token or None
+        """
+        token_url = self.API_URL.path.add('/v2/tokens/create/')
+        token_url.args['username'] = username
+        token_url.args['password'] = password
```
abought: In general we try to discourage sending passwords as part of the URL. For example, if someone read the network traffic (or our server logs), they'd have access to passwords in plaintext- even if we had mechanisms in place not to store passwords normally.

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,196 @@
...
+    def _find_or_create_user(self, username, password):
+        """ Checks to see if username exists in DB and compares password hashes.
+            Tries to authenticate and create user if a valid user/oauth_token is not found.
+
+        :return: user
+        """
+        try:
+            user = session.query(User).filter(User.osf_login == username).one()
+            assert user.oauth_token
+            assert user.password = generate_hash(password)
```
abought: In general, assertions should be for things that you never ever expect to happen- try to avoid using them for testing regular error conditions.

Also, this try/except block is checking for two different and unrelated kinds of errors; if possible try to keep as little stuff as possible in a given try block, and only test for one error condition at a time. (bad db query vs wrong password)

osfoffline/utils/authentication.py
mfraezz: https://github.com/CenterForOpenScience/osf.io/blob/develop/framework/auth/cas.py#L66

AFAIK, this is the only way to avoid a terribly messy situation involving pretending to be a web browser to CAS, faking a csrf-token, and ultimately using a cookie to make a request similar to this, only with the password in the request body. You can ask Michael about it if you want more details, but this approach is preferable.

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,189 @@
...
+        headers = {'content-type': 'application/json'}
+
+        try:
+            resp = yield from aiohttp.request(method='POST', url=token_url.url, headers=headers, data=token_request_body, auth=(username, password))
+        except (aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError, aiohttp.errors.TimeoutError):
+            # No internet connection
+            self.display_error('Unable to connect to server. Check your internet connection or try again later.')
+        except Exception as e:
+            # Invalid credentials probably, will be prompted later
+            logging.exception(e)
```
abought: The first argument to logging.exception is a message, not the actual exception object. (it will handle logging that on its own)

abought: Leaving aside the bare exception issue- why doesn't this clause call `self.display_error`, which was presumably intended as the reusable error handling method?

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,189 @@
...
+                return user
+
+        return
+
+    def display_error(self, message, fail=True, e=None):
+        """ Displays error message to user, sets fail flag unless otherwise specified,
+            and logs exception if given.
+        """
+        if e:
+            logging.exception(e)
```
abought: See note above about logging.exception, which should only be called from within an exception handler.
https://docs.python.org/3/library/logging.html#logging.Logger.exception

You can log any exception object in a standard `logging.<level>()` call (outside the exception handler) by passing in the `exc_info` argument:
https://docs.python.org/3/library/logging.html#logging.Logger.debug

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,189 @@
...
+            self.display_error('Username and password required for login.')
+            return
+        user = yield from self._find_or_create_user(username, password)
+        if self.failed_login:
+            return
+        if user:
+            user.logged_in = True
+            try:
+                save(session, user)
+            except Exception as e:
```
abought: Try to avoid bare exceptions when possible. Since we except this to catch database errors, an appropriate exception class might be:
`sqlalchemy.exc.SQLAlchemyError`
http://docs.sqlalchemy.org/en/rel_0_7/core/exceptions.html#sqlalchemy.exc.StatementError

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,189 @@
...
+    def _authenticate(self, username, password):
+        """ Tries to use standard auth to authenticate and create a personal access token through APIv2
+
+            :return: personal_access_token or None
+        """
+        token_url = furl.furl(settings.API_BASE)
+        token_url.path.add('/v2/tokens/')
+        token_request_body = '{"data": {"attributes": {"name": "OSF-Offline ' + \
+            str(datetime.date.today()) + \
+            '", "scopes": "osf.full_write"}, "type": "tokens"}}'
```
abought: I suggest moving the value of `scopes` into the settings file, to ease maintainability (especially if we decide in the future that the application should depend on a more restrictive set of scopes).

abought:
Pass will resume in the morning. I need to go code review the inside of my eyelids. :fallen_leaf:

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,189 @@
+import asyncio
+import datetime
+import furl
```
abought: Trivial FYI: furl is actually an external package, not part of the standard library

osfoffline/utils/authentication.py
mfraezz: It originally did, but when an exception was raised here that wasn't caught by the above except (manually raised), it prompted the user twice.

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,201 @@
...
+        """ Tries to use standard auth to authenticate and create a personal access token through APIv2
+
+            :return: personal_access_token or None
+        """
+        token_url = furl.furl(settings.API_BASE)
+        token_url.path.add('/v2/tokens/')
+        token_request_body = '{"data": {"attributes": {"name": "OSF-Offline ' + \
+            str(datetime.date.today()) + \
+            '", "scopes": "' + settings.APPLICATION_SCOPES + \
+            '"}, "type": "tokens"}}'
```
abought: Avoid building JSON via string concatenation- there are a variety of ways that could break. Instead, build a python dictionary and then pass `json.dumps(token_request_body)` in as the data argument of `aiohttp.request`

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,201 @@
...
+        except Exception as e:
+            # Invalid credentials probably, but it's difficult to tell
+            # Regadless, will be prompted later with dialogbox later
+            # TODO: narrow down possible exceptions here
+            logging.error(e.message)
+        else:
+            if not resp.status == 201:
+                return None
+            json_resp = yield from resp.json()
+            return json_resp['data']['attributes']['token_id']
```
abought: As discussed previously: we want to keep all API data parsing abstracted in one place later (we've already had to shift API formats twice, and centralizing logic eases refactoring pain). Currently the other parsers live in `polling_osf_manager/remote_objects.py`.

osfoffline/settings/app_settings.py
```diff
@@ -23,6 +24,10 @@
     API_BASE = 'https://staging-api.osf.io'
     FILE_BASE = 'https://staging-files.osf.io'
 
+# if PROD_MODE:  # TODO: uncomment these before release
```
abought: There variables aren't really needed. When we go to production, we'll just change the non-dev-mode variables to point to production directly.

(Production vs dev mode won't be configurable in the deployed executable)

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,201 @@
...
+        try:
+            resp = yield from aiohttp.request(method='POST', url=token_url.url, headers=headers, data=token_request_body, auth=(username, password))
+        except (aiohttp.errors.ClientTimeoutError, aiohttp.errors.ClientConnectionError, aiohttp.errors.TimeoutError):
+            # No internet connection
+            self.display_error('Unable to connect to server. Check your internet connection or try again later.')
+        except Exception as e:
+            # Invalid credentials probably, but it's difficult to tell
+            # Regadless, will be prompted later with dialogbox later
+            # TODO: narrow down possible exceptions here
+            logging.error(e.message)
```
abought: Why note use `logging.exception("Login failed")`? (should automatically capture the traceback)

osfoffline/utils/authentication.py
```diff
@@ -0,0 +1,201 @@
...
+                else:
+                    user.oauth_token = personal_access_token
+
+            if not self.failed_login:
+                return (yield from self._populate_user_data(user, username, password))
+            else:
+                return
+
+    @asyncio.coroutine
+    def _populate_user_data(self, user, username, password):
```
abought: Is it an intentional decision to not save the user object (created here) until line 175 of `log_in`?

(Asking because we had a different bug today where the lack of save was unintentional)

osfoffline/views/start_screen.py
```diff
@@ -35,73 +35,20 @@ def __init__(self):
...
+            user = yield from auth_client.log_in(username=username, password=password)
+        except Exception as e:
+            logging.exception(e)
+            user = None
 
+        if user:
+            logging.debug('Successfully logged in user: {}'.format(user))
             self.close()
+        else:
+            logging.debug('Login Failed')
```
abought: Separation of concerns note: arguably, `AuthClient` should not be responsible for UI work. It should raise errors and allow the UI client decide how to handle them with an appropriate message.

abought: Also, a login failure is something we want to actively record. `logging.debug` is best used for statements that are never of interest to the user (ran function, value at time x, etc)

In production mode, only messages of level INFO or above are saved to the log file.

osfoffline/views/rsc/gui/startscreen_gui/startscreen.ui
```diff
@@ -41,45 +41,50 @@
...
-      <width>141</width>
+      <x>30</x>
+      <y>50</y>
+      <width>110</width>
       <height>21</height>
      </rect>
     </property>
     <property name="text">
-     <string>Personal Access Token</string>
+     <string>Username (email):</string>
```
abought: To be consistent with the OSF homepage, just call this field "Email". (don't leave people guessing about whether a separate username is needed)

screen shot 2015-11-11 at 12 15 45 am

abought:
If I log out, then try to log in using a known username with the wrong password, I do not get an error message/ dialog box.

abought:
After logging in with the wrong password, then later entering a correct password, I get an error as follows:
```
[2015-11-11 00:18:40,940][MainThread][base_events.py][ERROR][asyncio]: Unclosed response
client_response: <ClientResponse(https://staging-api.osf.io/v2/tokens/) [401 UNAUTHORIZED]>
```

What does it mean for my response to be unclosed?

abought:
I deleted my personal access token via the website, then tried to log in again (to what should have been an existing user). I got the following message:
```
[2015-11-11 00:22:46,895][MainThread][start_screen.py][ERROR][root]: 'data'
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/views/start_screen.py", line 42, in log_in
    user = yield from auth_client.log_in(username=username, password=password)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/utils/authentication.py", line 169, in log_in
    user = yield from self._find_or_create_user(username, password)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/utils/authentication.py", line 124, in _find_or_create_user
    return (yield from self._populate_user_data(user, username, password))
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/utils/authentication.py", line 153, in _populate_user_data
    remote_user = RemoteUser(json_resp['data'])
KeyError: 'data'
```

abought:
Here are some acceptance tests I've outlined:

1. Log in to a fresh database with a new user
2. Without closing OSF offline, confirm that I can log out and immediately back in (same token used)
3. Log out of my account from OSF offline again. Verify that if I delete my personal access token, a new one is seamlessly created for use by the application
4. Change my password on the OSF. Verify that OSF offline does not continue to require my old password due to the password hashing mechanism

osfoffline/views/rsc/gui/startscreen_gui/startscreen.ui
abought: While you're in QT designer, could you also change the title of the dialog to something other than "startscreen"?

(Technically another ticket, but this will save time later) 

screen shot 2015-11-11 at 12 30 18 am

abought:
After logging out (via the preferences panel), still in the session described so far, I still see the OSF offline icon in the system tray. (and it still tells me I'm up to date, though whether I'm logged in is questionable at best)

But trying to reopen the preferences panel gives the error below (on the console):
```
[2015-11-11 00:31:12,495][MainThread][log_config.py][ERROR][root]: Uncaught exception: 
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/application/main.py", line 203, in open_preferences
    self.preferences.open_window(Preferences.GENERAL)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/views/preferences.py", line 108, in open_window
    self.selector(tab)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/views/preferences.py", line 113, in selector
  File "/Users/andyboughton/code/cos/OSF-Offline/env/lib/python3.5/site-packages/sqlalchemy/orm/query.py", line 2534, in one
    raise orm_exc.NoResultFound("No row was found for one()")
sqlalchemy.orm.exc.NoResultFound: No row was found for one()
```
screen shot 2015-11-11 at 12 33 35 am

abought:
In addition to some error logging and other notes, I'm seeing some distinct weird behaviors / potential bugs when I try to use this login flow. (see notes above) Will suspend review pending further discussion but probably won't be able to make the 1pm deploy Wednesday.

osfoffline/utils/authentication.py
mfraezz: Yes, lack of save until completed auth flow with nothing breaking was intentional.

osfoffline/utils/authentication.py
mfraezz: This is the only place where a response of this type will need to be parsed. Having a class to do only this seems excessive, especially when this information is tied to a (local, not remote) user.

PR #32, opened 2015-11-09
Alert user that they have not saved changes

ticklemepierce:
Purpose
-
Previously a user would not be notified that they had not saved their changes when select which projects should be synced and which to ignore.  This meant that a user may make a bunch of changes, close out of the window, and then be confused as to why nothing was being synced.  This PR seeks to address this issue by opening a window alerting the user that they have unsaved changes (if applicable)

screen shot 2015-11-09 at 2 42 45 pm

Changes
-
First I added a list field to the class to keep track of which projects were checked the last time the project was saved.  Then I bumped the code that creates guid_list out into its own method and called this inside of closeEvent and compared to the list of checked projects from the last save.  If they were equal, preferences closes normally, otherwise a message box pops up informing the user that they have unsaved changes.  They can either click a button to remain in preferences and review or save their changes or click a button to discard changes and close out of preferences.

Side effects
-
Due to the fact that this is predominately just a feature to improve UX there should not be any significant side effects.  There may be edge cases that cause the pop to show when it shouldn't or vice versa but the chances of this are pretty slim.

[#OSF-4409]


ticklemepierce:
Lol accidentally did this on master

PR #33, opened 2015-11-09
[Feature] Alert user that they have not saved changes

ticklemepierce:
Purpose
-
Previously a user would not be notified that they had not saved their changes when select which projects should be synced and which to ignore.  This meant that a user may make a bunch of changes, close out of the window, and then be confused as to why nothing was being synced.  This PR seeks to address this issue by opening a window alerting the user that they have unsaved changes (if applicable)

screen shot 2015-11-09 at 2 42 45 pm

Changes
-
First I added a list field to the class to keep track of which projects were checked the last time the project was saved.  Then I bumped the code that creates guid_list out into its own method and called this inside of closeEvent and compared to the list of checked projects from the last save.  If they were equal, preferences closes normally, otherwise a message box pops up informing the user that they have unsaved changes.  They can either click a button to remain in preferences and review or save their changes or click a button to discard changes and close out of preferences.

Side effects
-
Due to the fact that this is predominately just a feature to improve UX there should not be any significant side effects.  There may be edge cases that cause the pop to show when it shouldn't or vice versa but the chances of this are pretty slim.

[#OSF-4409]

### DIFF
--- a/osfoffline/views/preferences.py
+++ b/osfoffline/views/preferences.py
@@ -1,7 +1,7 @@
 import os
 import logging
 
-from PyQt5.QtWidgets import (QDialog, QFileDialog, QTreeWidgetItem)
+from PyQt5.QtWidgets import (QDialog, QFileDialog, QTreeWidgetItem, QMessageBox)
 from PyQt5.QtCore import QCoreApplication, Qt
 from PyQt5.QtCore import pyqtSignal
 from osfoffline.views.rsc.preferences_rc import Ui_Preferences  # REQUIRED FOR GUI
@@ -11,7 +11,6 @@ from osfoffline.polling_osf_manager.api_url_builder import api_url_for, NODES, U
 from osfoffline.polling_osf_manager.remote_objects import RemoteNode
 import requests
 import osfoffline.alerts as AlertHandler
-
 
 class Preferences(QDialog):
     """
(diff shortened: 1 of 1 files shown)


abought:
The button placement makes sense on MS Windows (our primary target audience), but typically button order is backwards on Mac (good action on the right, cancel action on the left)

Does QT provide any way to customize an ok/cancel type dialog box, so that the button placement is handled by QT in an OS-appropriate fashion?

osfoffline/views/preferences.py
```diff
@@ -36,12 +35,39 @@ def __init__(self):
 
         self.preferences_window.changeFolderButton_2.clicked.connect(self.update_sync_nodes)
         self.tree_items = []
+        self.checked_items = []
         self.setup_slots()
 
+    def get_guid_list(self):
+        guid_list = []
+        for tree_item in self.tree_items:
+            for name, id in [(node.name, node.id) for node in self.remote_top_level_nodes]:
```
abought: Just for future reference, try to avoid variable names that shadow builtins (like the function `id()`).

osfoffline/views/preferences.py
```diff
@@ -36,12 +35,39 @@ def __init__(self):
...
+                if name == tree_item.text(self.PROJECT_NAME_COLUMN):
```
abought: This function just gets the node associated with a project name. A note, but also a possible bug:

One way to boost performance would be to create a dictionary representation of `remote_top_level_names` before looping over tree items (to save iteration every time):
`node_lookup = {node.name : node.id for node in self.remote_top_level_nodes}`

BUT: what does this function do if two projects (with different node IDs) have the same name?

abought:
Looks good to me. There is a potential issue with duplicate names, but that is better handled under the scope of #OSF-4408. @ticklemepierce has some suggestions on how to fix this function as part of that ticket.

osfoffline/views/preferences.py
```diff
@@ -135,6 +156,7 @@ def create_tree_item_for_each_top_level_node(self):
 
             if node.id in user.guid_for_top_level_nodes_to_sync:
                 tree_item.setCheckState(self.PROJECT_SYNC_COLUMN, Qt.Checked)
+                self.checked_items.append(node.id)
```
chennan47: it needs to check if the node.id is already in the list. If you don't hit the update button, this list won't get assign as the guid_list and thus will append the node.id again make two lists different. Will fix this while I merge this pr.

PR #34, opened 2015-11-09
Fix DB error when two users share a computer

abought:
## Purpose
Fix an error reported by @nicipfeiffer  and @NatashaRichter in which trying to log out (then in as a different user) would give a database-related error message. ("Unable to save user data; please try again later")

## Summary of changes
Disable the `osf_login` field of the database model. It is intended to only contain an email address, and in the absence of that data, every single user was being given the same placeholder value. 

Since this violated the `unique` constraint on that field, users were seeing an error message.

## Side effects
Didn't update the tests- they were already broken, and likely will remain so until after this field is needed again. (we will begin collecting email addresses as part of #26 )


abought:
Whoops, wrong branch.

PR #39, opened 2015-11-10
Refactored preferences to no longer use node name as a dependancy

ticklemepierce:
Purpose
-
Previously there was a double for loop comparing tree item names with node names to link the two together.  This poses two problems:
 1) It is O(n^2) time when it could be O(n) and this is slow with large numbers of nodes.
 2) If two nodes have the same name it may link the wrong things

Changes
-
Changed tree_items to be a list of tuples so that the first tuple index is the tree_item and the second is the node id.  This allows us to do one loop and then get the part we need and they will always be linked together properly.

Side Effects
-
Should not be any because everything should be linked properly.

[#OSF-5152]

### DIFF
--- a/osfoffline/views/preferences.py
+++ b/osfoffline/views/preferences.py
@@ -41,10 +41,8 @@ class Preferences(QDialog):
     def get_guid_list(self):
         guid_list = []
         for tree_item in self.tree_items:
-            for name, id in [(node.name, node.id) for node in self.remote_top_level_nodes]:
-                if name == tree_item.text(self.PROJECT_NAME_COLUMN):
-                    if tree_item.checkState(self.PROJECT_SYNC_COLUMN) == Qt.Checked:
-                        guid_list.append(id)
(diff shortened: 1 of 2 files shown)


osfoffline/views/preferences.py
```diff
@@ -41,10 +41,8 @@ def __init__(self):
     def get_guid_list(self):
         guid_list = []
         for tree_item in self.tree_items:
```
abought: To keep this readable/maintainable, please use tuple unpacking here:

```
for tree_item, node_id in self.tree_items:
    if tree_item.checkState(self.PROJECT_SYNC_COLUMN) == Qt.Checked:
        guid_list.append(node_id)
```

abought:
Before this merges, can I request that we also change the display of node names in the dialog, to match whatever `title + node ID` format we decide to use for folder names?

Otherwise, users won't have any way of knowing which node ID they selected. (even though the tree will now be more accurate)

abought:
Code review done. Thanks!

PR #41, opened 2015-11-10
[Feature][Fix] unique project folder names

mfraezz:
Purpose
======
On the OSF, you can have two projects with the same name. In a filesystem, having two directories of the same name in the same location doesn't work as well. 

Changes
=======
Append node/folder ID from the OSF to the name when creating a folder path

Side Effects
=========
* Improved logging for a `todo handle exception`.
* None found during testing. (But other unrelated bugs uncovered during testing)

[#OSF-4408]

### DIFF
--- a/osfoffline/database_manager/models.py
+++ b/osfoffline/database_manager/models.py
@@ -9,6 +9,7 @@ from sqlalchemy.orm import relationship, backref, validates
 from sqlalchemy import Column, Integer, Boolean, String, DateTime
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
+from osfoffline.utils.path import make_folder_name
 
 Base = declarative_base()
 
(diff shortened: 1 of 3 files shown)


osfoffline/database_manager/models.py
```diff
@@ -100,10 +101,9 @@ def path(self):
         # +os.path.sep+ instead of os.path.join: http://stackoverflow.com/a/14504695
 
         if self.parent:
-            return os.path.join(self.parent.path, 'Components', self.title)
+            return os.path.join(self.parent.path, 'Components', make_folder_name(self.title, self.osf_id))
```
abought: According to your own function definition, the second argument is a kwarg, not positional- update to match.

abought:
Ok for merge- using the DB path function seems to hook in cleanly. Less code needed = good.

There are some small changes that can be made post-merge; we'll do further verification when we regression test in preparation for the deployment.

PR #42, opened 2015-11-10
[Feature] Don't delete on modify

mfraezz:
Purpose
======
Close previously unknown bug. When a file is modified, the OS typically uses temp files. In many cases, this causes what should be viewed as a modify event to be a delete event, presumably because the original file is deleted and the temp file is renamed to the original file in one action.

Changes
=======
On delete, sleep for a second to make sure any filesystem changes are finished, then check to see if the file still exists on disk. If it does, don't delete it remotely.

Side Effects
=========
None

### DIFF
--- a/osfoffline/filesystem_manager/osf_event_handler.py
+++ b/osfoffline/filesystem_manager/osf_event_handler.py
@@ -4,6 +4,7 @@ storing the data into the db, and then sending a request to the remote server.
 """
 import asyncio
 import logging
+import os
 
 from watchdog.events import FileSystemEventHandler, DirModifiedEvent, DirCreatedEvent, FileCreatedEvent
 from osfoffline.database_manager.models import Node, File, User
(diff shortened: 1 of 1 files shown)


osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -203,18 +204,18 @@ def on_deleted(self, event):
...
-        save(session, item)
-
-        logging.info('{} set to be deleted'.format(src_path.full_path))
+        # put item in delete state after waiting a second and
+        # checking to make sure the file was actually deleted
+        yield from asyncio.sleep(1)
+        if not os.path.exists(item.path):
+            item.locally_deleted = True
+            # nodes cannot be deleted online. THUS, delete it inside database. It will be recreated locally.
+            if isinstance(item, Node):
```
abought: Note for regression testing: check whether files inside a project are correctly recreated as well. (there may be DB conflicts when files with the same ID are re-downloaded)

Presumably, there would be no DB cruft because all DB entries referencing files inside the deleted project folder will be cleared from the DB (because the nodes.files relationship is set to cascade deletes).

However, SQLAlchemy docs suggest that may not be enough:
http://docs.sqlalchemy.org/en/rel_0_9/orm/cascades.html
```
The all symbol is a synonym for save-update, merge, refresh-expire, expunge, delete, and using it in conjunction with delete-orphan indicates that the child object should follow along with its parent in all cases, and be deleted once it is no longer associated with that parent.
```

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -203,18 +204,18 @@ def on_deleted(self, event):
...
+                session.delete(item)
+                save(session)
```
abought: Please wrap all DB save calls in a try/except block, to avoid an unhandled DB exception.

(as we saw yesterday, data corruption does happen- we don't want it to hang the client)

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -203,18 +206,25 @@ def on_deleted(self, event):
...
+        yield from asyncio.sleep(1)
+        if not os.path.exists(item.path):
+            item.locally_deleted = True
+            # nodes cannot be deleted online. THUS, delete it inside database. It will be recreated locally.
+            if isinstance(item, Node):
+                session.delete(item)
+                try:
+                    save(session)
+                except SQLAlchemyError as e:
+                    logging.warning(e.message)
```
abought: Note to Nan: postmerge, replace this line and line 225 with:

`logging.exception('Error deleting node from database')`

DB errors should never happen, so we want to capture a more detailed traceback (which logging.exception will do automatically) in addition to just the message.

abought:
Good to merge. Two requests for cleanup after merge, to improve logging of an edge case DB issue.

PR #46, opened 2015-11-10
[hold][Feature] Cross platform: halt additional instances

mfraezz:
Purpose
=======
Allow only one instance of OSFO to run

Changes
=======
Adapt `SingleInstance` class from `tendo` (only supports Python 2.6-3.2, adaptation supports current)

Side Effects
=========
None

Note
====
Tested on OS X and Windows

[#OSF-5149]


osfoffline/utils/singleton.py
```diff
@@ -0,0 +1,69 @@
...
+                self.fd = os.open(
+                    self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
+            except OSError:
+                type, e, tb = sys.exc_info()
+                if e.errno == 13:
+                    logging.error("Another instance is already running, quitting.")
+                    if callback:
+                        callback()
+                    sys.exit(-1)
+                print(e.errno)
```
chennan47: This print statement doesn't seem necessary.

chennan47:
CR1 done. Looks good. Verified it works for both mac and windows system.

abought:
I will pause on reviewing this one until after tomorrow's deploy deadline.

PR #48, opened 2015-11-11
Code/Logic clean up

chrisseto:
Cleans up the OSF polling class
Fixes ~50% of the flake8 issues
Cleans up the requirements file
Attempts to consolidate the logic for the background worker.

ping @abought

Notes:
Seems to work locally, ran into an error with a file being a folder or visa versa, may be an existing issue.
Uncaught exceptions will be logged and will stop the polling functionality but will not stop the loop
If you see `CancelledException`s, dont worry, it's expected.


brianjgeiger:
It appears that there are some merge conflicts. I know you will be surprised about this based on the scope of the changes.

brianjgeiger:
I sat down with @chrisseto for an in-person code review. There's a couple of minor things that will get changed in the next commit (mostly converting logging statements to logger statements), and it should be good. As near as I can tell, this should take care of a good deal of weird instability issues. But it's a big change to fundamental app logic, so everything will have to be tested as if it's brand new. This will especially affect issues that are hard to reproduce because timing is involved.

chrisseto:
Refer to https://github.com/CenterForOpenScience/OSF-Offline/pull/54

abought:
This PR has fallen into the infinite void between universes, so I'm closing it in favor of the aforementioned #54.

PR #50, opened 2015-11-11
[hold][Feature] Error handling around creation and renaming of files & folders

ticklemepierce:
Purpose
-
Previously when a user renamed a node folder it would mess up syncing for various DB related issues.  This fixes that error and gives the user an alert telling them renaming is bad.  There are also errors for attempting to rename the Components folder.  This code runs prior to other error handling code and is thus able to prevent any other error handling code from running if the user does something that they shouldn't.

Changes
-
Adding extra error handling and checks in the is_illegal method of the osf_event_handler.  These checks allow us to ignore changes on files in the case that the node was renamed and handle the error when an event is called for the node itself.  This is due to the fact that watchdog acts recursively but the recursion seems to go children first, and on_moved is therefore called on all of the children of the renamed node prior to be called on the node itself.  This method is extensible and allows us to add many checks and then halt the rest of the error handling.

Side Effects
-
There should not be any side effects from this, although there is the possibility of getting into a weird state that may inhibit syncing, although it is not clear what would cause this.  

Notes
-
The reason I had to combine into one method for now is that having two different ones was making it so that if I had an illegal rename (i.e. renaming a node) I would get an alert saying that I couldn't create something called 'Components', when I didn't (it just thinks I did based on the multiple events that fire).  In is_illegal, no matter whether its a folder or file, I check for illegal events above it (I.E. a node rename( and do nothing in this case knowing that it will be handled later.  If this is not the case, I can easily check for what did trigger the event.

What works
-
 * Renaming a node
  * Will provide an alert informing the user that this is not what they should do
  * If they then change it back to normal, life is great and nothing weird happens
  * If they then change it to something else they get an alert asking them to please change it to its proper name
  * If they don't change it to it's proper name, the next time that everything is synced, a new folder for the project will be created
 * Renaming components
  * Tells the user that they should not rename this folder and that it will mess things up.
 * Adding something named components
  * Tells them they should not create things named 'Components' as it is a reserved word

What doesn't work
-
 * Renaming the OSF folder does not tell the user anything
 * Renaming the base folder does not tell the user anything

[#OSF-5154]


osfoffline/settings/defaults.py
```diff
@@ -16,7 +16,7 @@
 POLL_DELAY = 5 * 60  # seconds
 
 # Time to keep alert messages on screen (in milliseconds); may not be configurable on all platforms
-ALERT_TIME = 1000  # ms
+ALERT_TIME = 10000  # ms
```
abought: I'm guessing this is a remnant of local debugging efforts, rather than something intentionally committed?

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -67,7 +67,9 @@ def on_moved(self, event):
         item = self._get_item_by_path(src_path)
 
         if isinstance(item, Node):
-            AlertHandler.warn('Cannot manipulate components locally. {} will stop syncing'.format(item.title))
+            if item.title != dest_path.name:
+                AlertHandler.warn('Cannot manipulate components locally. {} will stop syncing'.format(item.title))
+                # os.rename(str(dest_path), item.path)
```
abought: Remove this line.

(And for future reference: If you were moving files/folders- in a program that had far more control over the user's behavior than osf offline does- you'd probably want to look into something like `shutil.move` instead)

abought:
This is marked WIP, so I'll hold off on reviewing for now. Let me know if this is expected to be part of Tuesday's release.

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -204,6 +204,7 @@ def on_modified(self, event):
         """
 
         if isinstance(event, DirModifiedEvent):
+            # AlertHandler.warn('Modifying folders is not recommended')
```
abought: Why is this commented out?

ticklemepierce: This will run every time any change is made to any folder.  We don't want this.  I'll push up a commit without it if you want.  TBH totally forgot about this change.

osfoffline/settings/defaults.py
```diff
@@ -17,7 +17,7 @@
 POLL_DELAY = 5 * 60  # seconds (set to 1 min now for QA purpose, #todo change it back)
 
 # Time to keep alert messages on screen (in milliseconds); may not be configurable on all platforms
-ALERT_TIME = 1000  # ms
+ALERT_TIME = 10000  # ms
```
abought: Please change this back to 1 second for now.

We display a lot of alerts. With this change, each individual alert appears for 10 seconds (during which interval many alerts drop out of the queue and aren't displayed at all)

abought:
Earlier you commented about SQLAlchemy errors, in which the state of the database could become corrupted in certain cases. Can you please point me to where that was resolved?

(This is a big issue to resolve before the PR gets merged)

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -325,13 +329,110 @@ def _get_item_by_path(self, path):
...
-                return False
-        except Exception:
-            pass  # Similar path-related error will occur and be caught later, better error message will be displayed then
+                item = self._get_item_by_path(src_path)
+            except ItemNotInDB:
+                # Either is the components folder or something that was previously illegally named
+                try:
+                    # Item is the components folder so lets see if there is more stuff above it
+                    item = self._get_item_by_path(src_path.parent)
+                    # Fixme: probably could use another assert statement somewhere
```
abought: Is that so?

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -325,13 +329,110 @@ def _get_item_by_path(self, path):
...
+                            return True
+                        if curr.top_level:
+                            running = False
+                            return False
+                        else:
+                            curr = curr.node
+                    raise
+            if dest_path.name in reserved_words:
+                AlertHandler.warn('Cannot have a custom file or folder named {}'.format(dest_path.name))
+                print('Cannot have a custom file or folder named {}'.format(dest_path.name))
```
abought: Please remove all print statements before pushing forward to review,

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -325,13 +329,110 @@ def _get_item_by_path(self, path):
...
+                            AlertHandler.warn('{} will not sync until it is renamed properly'.format(dest_path.name))
+                            print('{} will not sync until it is renamed properly'.format(dest_path.name))
+                        return True
+            try:
+                new_parent_item = self._get_parent_item_from_path(dest_path)
+            except ItemNotInDB:
+                if isinstance(item, Node):
+                    if make_folder_name(item.title, item.osf_id) != dest_path.name and parent == False:
+                        AlertHandler.warn('Cannot manipulate components locally. {} will stop syncing'.format(item.title))
+                        print('Cannot manipulate components locally. {} will stop syncing'.format(item.title))
```
abought: Please remove print statements before pushing to review

abought:
I'm going to put a hold on this PR for tonight's merge.

While I'd love to see a bugfix for this go in, this code is very convoluted- there are multiple nested try/except blocks (including one four levels deep), not to mention various if/else blocks and while loops. If this introduces a bug in mroning QA, I'm not convinced that we'd be able to refactor and reason about this code in time to land a fix. 

We'll discuss some positive cleanup steps when the immediate rush is past, but for now I'll hold off merging it in to develop.

abought:
Also, can you link to an OSF ticket for this issue?

abought:

PR #54, opened 2015-11-11
Feature/refactor

chrisseto:


start.py
```diff
@@ -39,10 +40,11 @@ def start():
     QApplication.setQuitOnLastWindowClosed(False)
 
     osf = OSFApp()
+
     osf.start()
 
     osf.hide()
-    app.exec_()
+    sys.exit(app.exec_())
```
abought: Good!

osfoffline/settings/defaults.py
```diff
@@ -0,0 +1,29 @@
+# Just to insure requirement
+import colorlog  # noqa
+
+# Development mode: use a local OSF dev version and more granular logging
+DEV_MODE = False  # TODO (abought): auto-set flag when using `inv start_for_tests`
```
abought: Deprecated setting

osfoffline/views/preferences.py
```diff
@@ -197,16 +223,5 @@ def get_remote_top_level_nodes(self):
         except Exception as e:
             logging.warning(e)
 
+        self.finished.emit(remote_top_level_nodes)
```
abought: If the server connection fails for any reason at all, the user will not be notified that the fetch failed- the prefs panel will simply show an empty list. We should improve how fetch failures are communicated to the user.

osfoffline/polling_osf_manager/remote_objects.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, remote_dict):
     def validate(self):
         super().validate()
         assert self.provider
-        assert self.move_url if not self.id  else True
+        assert self.move_url if not self.id else True
```
abought: Am reviewing this a bit backwards to divide-and-conquer.

Done reviewing all items below this line.

(Arguably we're misusing asserts here in potentially freeze-inducing ways, but I won't hold that here since this is only cleanup to existing code)

abought:
As provided, I can not run the code in this branch following the settings file reorg.

`OSF-Offline/osfoffline/utils/__init__.py` is still trying to import `from osfoffline.settings import log_config`, which is no longer valid.

osfoffline/utils/__init__.py
```diff
@@ -5,18 +5,18 @@
 import logging.config
 import logging.handlers
 import os
-import sys
```
abought: Please add this back. This removal breaks the custom `excepthook` on line 22.

(sys.excepthook is used as a quick-and-dirty way to ensure that all uncaught exceptions get logged)

osfoffline/settings/log_config.py
```diff
@@ -1,70 +0,0 @@
...
-        }
-    },
-    'root': {
-        'level': log_level,
-        'handlers': ['console', 'logfile']
-    }
-}
-
-
-def capture_exceptions(exc_type, exc_value, tb):
```
abought: This function was removed entirely; it should be defined somewhere (perhaps in the same module as `start_app_logging`).

Without it the custom excepthook will break.

abought:
Since we're defining a filename-specific logger *and* including the filename in log file format string, there's a bit of redundancy (cosmetic):
`[Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]`

osfoffline/application/background.py
```diff
@@ -1,103 +1,82 @@
...
 
 
 class BackgroundWorker(threading.Thread):
+
     def __init__(self):
         super().__init__()
+
+        # Start out with null variables for NoneType errors rather than Attribute Errors
         self.user = None
-        self.osf_folder = ''
```
abought: This change should help with a certain rather nasty bug that occurred when the path was no longer valid (and it just tried to save to `/OSF` instead); thanks.

osfoffline/application/background.py
```diff
@@ -1,103 +1,82 @@
...
-            self.poller.stop()
+        logging.debug('Starting OSF polling')
+        self.start_osf_poller()
 
-    # todo: can refactor this code out to somewhere
+        logging.debug('Starting background event loop')
+        try:
+            self.loop.run_forever()
+        except Exception as e:
+            print(e)
```
abought: Use `logging.exception('Background worker failed unexpectedly')` instead- if the polling fails this hard, we should make sure to capture the traceback.

(are there circumstances where an uncaught exception stopping the loop would be normal? The bare `except Exception as e` is a bit vague wrt expected failure case)

osfoffline/application/background.py
```diff
@@ -114,32 +93,47 @@ def start_observing_osf_folder(self):
 
         try:
             self.observer.start()  # start
-        except OSError:
+        except OSError as e:
             # FIXME: Document these limits and provide better user notification.
             #    See http://pythonhosted.org/watchdog/installation.html for limits.
-            logging.warning('limit of watched items reached')
+            raise RuntimeError('Limit of watched items reached') from e
```
abought: Hadn't know python3 had `raise from`. Always fun to learn a new trick, however small. :)

osfoffline/application/main.py
```diff
@@ -127,55 +131,36 @@ def start(self):
...
             user = session.query(User).filter(User.logged_in).one()
             save(session, user)
             session.close()
-
-            QApplication.instance().quit()
-        except Exception as e:
-            # FIXME: Address this issue
-            logging.warning('quit broke. stopping anyways. Exception was {}'.format(e))
-            # quit() stops gui and then quits application
+        finally:
```
abought: This is a `try/finally` that doesn't appear to handle the exception; it's not clear to me that line 164 would be reached, in which case the log message may be a tad optimistic. Is something in this catchall fails, wouldn't the application still appear to hang and require a force-quit?

abought: Ah, nevermind- the quit is still inside the finally block. Mea culpa.

osfoffline/utils/__init__.py
chrisseto: sys.excepthook was breaking a lot of things. Uncaught exceptions will be logged anyways

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -17,10 +17,11 @@
 
 
 class OSFQuery(object):
-    def __init__(self, loop, oauth_token):
+    def __init__(self, loop, oauth_token, limit=5):
```
abought: :smile_cat: 

Since we presently use 8 workers in production AFAIK, perhaps a more reasonable limit would be 4? It's a small thing, but two people syncing at the same time could saturate the server wherein one person would be getting better service (more requests) than the other at a given time.

osfoffline/polling_osf_manager/osf_query.py
```diff
@@ -17,10 +17,11 @@
...
         self.headers = {
             'Authorization': 'Bearer {}'.format(oauth_token),
         }
+        self.throttler = asyncio.Semaphore(limit)
```
abought: Any reason this isn't a `BoundedSemaphore`? (genuinely asking)

chrisseto: No reason for it to be. The additional functionality provided `BoundedSemaphore` by is pointless here, we only care about upper limits.

osfoffline/polling_osf_manager/polling.py
```diff
@@ -29,42 +31,84 @@ def __init__(self, user, loop):
...
+            raise self.process_job.exception()
 
-        self._keep_running = False
+        # Other wise cancel them, the event loop will clean up for us
+        self.poll_job.cancel()
+        self.process_job.cancel()
         self.osf_query.close()
+        logger.info('OSF polling requested stopped.')
+
+    def handle_exception(self, future):
```
abought: I believe that futures are the future. (Let's be honest here; the children had their chance) 

This does appear to simplify a lot of the request cleanup; we'll regression test this heavily of course.

osfoffline/polling_osf_manager/polling.py
```diff
@@ -29,42 +31,84 @@ def __init__(self, user, loop):
...
 
-        self._loop.call_soon(
-            asyncio.async,
-            self.check_osf(remote_user)
-        )
+        self.process_job = asyncio.ensure_future(self.process_queue())
+        self.poll_job = asyncio.ensure_future(self.check_osf(remote_user))
+
+        self.poll_job.add_done_callback(self.handle_exception)
+        self.process_job.add_done_callback(self.handle_exception)
```
abought: `handle_exception` is a callback that mostly seems suited to dealing with failed requests.

But this same callback is also used to handle the task queue (`self.queue`), and the tasks on said on said queue can raise a bunch of exceptions that aren't dealt with in `handle_exception`. How would the exception handler deal with, for example, a process job item that failed to create a folder due to an OSError? Would it ever reach the cleanup/ cancel statements?

osfoffline/polling_osf_manager/polling.py
```diff
@@ -29,42 +31,84 @@ def __init__(self, user, loop):
...
-                future.set_result(resp['data'])
-                break
-            except (concurrent.futures._base.TimeoutError, aiohttp.errors.ClientTimeoutError):
-                AlertHandler.warn("Bad Internet Connection")
+                return (yield from self.osf_query.make_request(url, get_json=True))['data']
+            except aiohttp.ClientError:
+                pass
+            sleep_time += 5
+            logger.error('Unable to fetch user from OSF, sleeping for {} seconds'.format(sleep_time))
+            yield from asyncio.sleep(sleep_time)
```
abought: If I'm reading this correctly: for each time a server request fails, it will try again slightly later, indefinitely?

Should this be bounded on the upper end (eg guarantee it will retry no more than every hour)?

osfoffline/utils/__init__.py
abought: interesting. Can you give an example of what it was breaking?

osfoffline/polling_osf_manager/polling.py
chrisseto: Yes, any exception raised in the `process_queue` or `check_osf` will be handled by `handle_exception`

osfoffline/polling_osf_manager/polling.py
```diff
@@ -130,54 +174,37 @@ def make_local_remote_tuple_list(self, local_list, remote_list):
...
-            except (aiohttp.errors.ClientConnectionError, aiohttp.errors.ClientTimeoutError,
-                    concurrent.futures._base.TimeoutError):
-                # NOTE: can't work with partial list! That would suggest that nodes were created online.
-                AlertHandler.warn("Bad Internet Connection")
-                # waits till the end of a sleep to stop. thus can make numerous smaller sleeps
-                for i in range(POLL_DELAY):
-                    yield from asyncio.sleep(1)
-                continue
-
+        while True:
```
abought: Just to be sure: previously we relied on `keep_running` to stop this loop, but now we use the future to cancel/cleanup instead?

chrisseto: Yup

abought:
After syncing a project initially (and then not touching the projects further), I see `CreateFolder` events being added to the queue every 5 min.

These events do not create folders successfully (I would see alert messages had the folders not already existed), but the fact they are continuously added to the `process_job` queue at all is a bit problematic. Is this a polling issue?

```
[2015-11-11 22:33:47,751][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Begining OSF poll
[2015-11-11 22:33:49,874][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Running <osfoffline.polling_osf_manager.polling_events.CreateFolder object at 0x11b771ba8>
[2015-11-11 22:33:51,128][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Running <osfoffline.polling_osf_manager.polling_events.CreateFolder object at 0x121b4ae48>
[2015-11-11 22:38:52,484][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Begining OSF poll
[2015-11-11 22:38:54,447][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Running <osfoffline.polling_osf_manager.polling_events.CreateFolder object at 0x1221f03c8>
[2015-11-11 22:38:55,171][Thread-8][polling.py][INFO][osfoffline.polling_osf_manager.polling]: Running <osfoffline.polling_osf_manager.polling_events.CreateFolder object at 0x1221f0d30>
```

osfoffline/polling_osf_manager/osf_query.py
abought: > No reason for it to be. The additional functionality provided BoundedSemaphore by is pointless here, we only care about upper limits.

Hmm. From the description, I assumed that upper (not just lower) limits were precisely what `BoundedSemaphore` would provide. Will look into that when I'm less tired then. :)

abought:
Pausing review for now, because it's midnight again. Will resume in the AM. Lots of general improvements, maybe some small new quirks.

Surplus event items are one thing that stands out, as well as some issues involving exception handling (and, separately, logging) Either way there are a few things we can clean up post-merge, provided we resolve a few small items that could be blockers for initial QA regression testing.

osfoffline/polling_osf_manager/polling.py
brianjgeiger: Yeah, I went back and forth on that because it's only for the login. It certainly wouldn't hurt, but I think the real solution will be a bit more complicated. If someone keeps trying to log in for weeks and suddenly gets internet, I'm expecting that the difference between waiting an hour and a week is negligible. 

I'd either bound it smaller, like 5 minutes, or put in a todo to have a better way of determining, cross-platform, that we have internet and not even bother trying until internet is good. I know that MacOS and iOS have "are you connected" hooks but I don't know about windows and linux.

abought: Broadly I agree. Two issues brought up during QA would help:
1. A manual "sync now" button that restarts polling on the spot, and,
2. Instead of just saying "up to date", the systray icon menu line that indicates status should say the last updated time (so people know when the last sync actually happened)

abought:
Will merge now so that we can build on it from there. Confirmed that `sys.excepthook` has problems playing nicely with threads ([possible reference](https://bugs.python.org/issue1230540) ), so will add followup PR to log exceptions.

Once that's done we can do a deploy to QA for light testing before going further.

PR #55, opened 2015-11-12
[Feature][discuss] Reindex DB when restarting observer...

mfraezz:
...so that any changes to the local filesystem (e.g. file creation, modification, deletion) during non-observed time are tracked.

[#OSF-5122]

Changes
=======
`LocalDBSync` class update for use, and used

Side Effects
=========
Because the tracked portion of the filesystem is walked through to check for changes, a drop in performance may occur when observer is reinitialized for very large sets of tracked files.


osfoffline/application/main.py
```diff
@@ -163,6 +165,14 @@ def quit(self):
             logger.info('Quitting application')
             QApplication.instance().quit()
 
+    def sync_now(self):
+        if not self.background_worker:
+            self.start()
+        if self.background_worker.is_alive():
```
abought: As discussed on another PR, `self.background_worker.is_alive` already seems to be called in `self.pause`.

Also, it seems there might be a race condition if the background worker stops in between the (first/second) check and the stop action. (EAFP might be a better way to go)

abought:
Can you suggest some test cases?

In a conversation with Himanshu (https://openscience.atlassian.net/browse/OSF-5163 ), he mentioned that there may be some hash-related issues with `sync_local_file_system_and_db`. I'd like to be sure that just turning it on is sufficient to actually fix these issues.

If hashes are needed, they are now (or will soon be) available on staging via waterbutler (but not yet apiv2 files endpoints; @brianjgeiger is working on that).

mfraezz:
Test cases include but are not limited to - 
After syncing one or more projects:
```
1. Open `Preferences`/`Settings`, add a new file to the project folder, close settings
2. Open `Settings`, edit existing file, close settings
3. Open `Settings`, delete existing file, close settings
4. Repeat 1-3 for remote files on the OSF
5. All possible combinations of 1-4
```
Check to ensure changes made during lapse in polling persist between local filesystem and the OSF

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -167,6 +136,12 @@ def _determine_event_type(self, local, db):
             raise TypeError
 
         event = None
+
+        if local:
+            # TODO: find better way of ignoring files we want to ignore
+            if local.name.startswith('.') or local.name.endswith('~'):
```
abought: In the polling class, we silently exclude the `.Ds_Store` file from uploads (though it still generates a user message). We should start abstracting all "bad filename" checks to a single centralized utility function.

https://github.com/abought/OSF-Offline/blob/develop/osfoffline/polling_osf_manager/polling.py#L171-171

mfraezz: That's what the `TODO` is for. Admittedly, not as descriptive as possible.

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -46,42 +43,21 @@ def _make_local_db_tuple_list(self, local, db):
...
+            assert isinstance(child, ProperPath)
+            local_children[child.full_path] = child
+
+        for child in self._get_children(db):
+            assert isinstance(child, Base)
+            db_children[self._get_proper_path(child).full_path] = child
+
+        return [
+            (local_children.get(path), db_children.get(path))
+            for path in set(list(local_children.keys()) + list(db_children.keys()))
```
abought: Not a required change, but one useful trick- instead of this creating two temporary lists and concatenating into a third temporary list, you can use `set(itertools.chain(d1.keys(), d2.keys()))`.

https://docs.python.org/3.5/library/itertools.html#itertools.chain

abought:
Note to reviewers: this (effectively) stops automatic syncing of files. Syncing will now occur only when the user manually clicks "sync now".

osfoffline/application/background.py
```diff
@@ -89,7 +91,7 @@ def start_folder_observer(self):
         self.observer = Observer()  # create observer. watched for events on files.
         # attach event handler to observed events. make observer recursive
         self.observer.schedule(self.event_handler, self.osf_folder, recursive=True)
-        # LocalDBSync(self.user.osf_local_folder_path, self.observer, self.user).emit_new_events()
+        LocalDBSync(self.user.osf_local_folder_path, self.observer, self.user).emit_new_events()
```
abought: If I'm reading this right, this doesn't just re-index when we restart the application. It also does a full re-index every time we click sync now, and every time the preferences panel is opened/closed. Thus turning this on will require regression testing of *all* sync events, not just those that happen while the application is closed.

This line appears in `start_folder_observer(...)`, which is called in `BackgroundWorker.run(...)`, which is called when the thread is started. Eg:
1. `OsfApp.start` (which is our intended goal)
2. `OsfApp.resume` (prefs panel close, syc now, etc)

abought:
Review pass done. I'll do some regression testing this evening before moving this forward to a merge. (I'd like to look into how `LocalDbSync` makes decisions, since it seems it will be inserted into so many places)

osfoffline/application/background.py
mfraezz: You are reading that right.

abought:
So, I’m still on the fence on #55- it inserts an untested new chunk of code into every sync operation, and even on a first glance the new file contains patterns we’ve found to be buggy in the past (https://github.com/abought/OSF-Offline/blob/develop/osfoffline/filesystem_manager/sync_local_filesytem_and_db.py#L55-55). I'm not 100% convinced that we could land this change without introducing side effects in the time available.


Instead of merging it outright (even though I love the potential gains), how would we feel about preparing a special release branch for the deploy to QA?

- Advantage: it would let QA test it, but we would be able to revert later if needed
- Disadvantage: it would be harder to build on top of this branch because it wouldn’t get merged directly into develop. 
  - This branch contains other desired functionality like Sync Now.

mfraezz:
The *old* file contain**ed** buggy patterns, and I fixed that (approximately) the same way that I fixed the original buggy pattern. You linked to the outdated version of this file on develop, not the updated one in this PR. All this file does is walk through the watched `OSF` directory and compare what's actually there with what the database expects, and handle any differences by updating the database and scheduling events. It's like `polling.py`, but simpler because it's not done in real time.

I think it'd be okay if we gave QA two separate distributables, but the one with this in it should receive the most attention, because this contains some functionality that we originally stated was necessary for the alpha release (e.g. tracking file changes whether OSFO is polling or not).

abought:
> The old file contained buggy patterns, and I fixed that (approximately) the same way that I fixed the original buggy pattern. You linked to the outdated version of this file on develop, not the updated one in this PR

Confirmed; I switched between way too many branches. Sorry about that.

We'll stick with the release-branch strategy. (even doing just one distribution would be ok) The goal isn't to prevent this from going out, but rather to give it a solid round of extra testing before we make it official and merge it in.

chennan47:
Already merge in https://github.com/CenterForOpenScience/OSF-Offline/tree/release for Monday deploy

PR #56, opened 2015-11-12
Preferences panel UI Changes

ticklemepierce:
Purpose
-
 * Renamed 'Preferences' to 'Settings'
 * Changed 'Update' button to 'Apply Settings'
 * Made the 'Apply Settings' button close the 'Settings' window
 * Fixed issued where the loading cursor would stay forever the second time you opened the settings window

Changes
-
 * Basic renaming changes in preferences.ui/preferences_rc.py using QT Creator.
 * In preferences.py added a call to close the window when the 'Apply Settings' button is clicked.
 * Moved around some code and added a check to see whether the tree has been populated yet or not to handle the cursor error

Side Effects
-
Due to the fact that these are all UI changes basically there should be no side effects.

[#OSF-5167]

### DIFF
--- a/osfoffline/views/preferences.py
+++ b/osfoffline/views/preferences.py
@@ -155,8 +155,13 @@ class Preferences(QDialog):
             user = session.query(User).filter(User.logged_in).one()
             self.preferences_window.label.setText(self._translate("Preferences", user.full_name))
 
-            self.preferences_window.treeWidget.setCursor(QtCore.Qt.BusyCursor)
-            self.node_fetcher.finished[list].connect(self.populate_item_tree)
+            self._executor = QtCore.QThread()
+            self.node_fetcher = NodeFetcher()
+            if len(self.tree_items) == 0:
+                self.preferences_window.treeWidget.setCursor(QtCore.Qt.BusyCursor)
+                self.node_fetcher.finished[list].connect(self.populate_item_tree)
+            else:
+                self.node_fetcher.finished[list].connect(self.check_tree_changed)
             self.node_fetcher.moveToThread(self._executor)
             self._executor.started.connect(self.node_fetcher.fetch)
             self._executor.start()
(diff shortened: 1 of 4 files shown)


osfoffline/views/preferences.py
```diff
@@ -154,8 +155,13 @@ def selector(self, selected_index):
             user = session.query(User).filter(User.logged_in).one()
             self.preferences_window.label.setText(self._translate("Preferences", user.full_name))
 
-            self.preferences_window.treeWidget.setCursor(QtCore.Qt.BusyCursor)
-            self.node_fetcher.finished[list].connect(self.populate_item_tree)
+            self._executor = QtCore.QThread()
```
abought: Why is this redefined/duplicated from L55-56?

osfoffline/views/preferences.py
```diff
@@ -164,6 +170,11 @@ def reset_tree_widget(self):
         self.tree_items.clear()
         self.preferences_window.treeWidget.clear()
 
+    def check_tree_changed(self, nodes):
+        if nodes != self.tree_items:
```
abought: If I read this correctly, `nodes` is a list of `RemoteNode` objects (see L207) , but `tree_items` is a list of `(tree_item, node.id)` tuples. Are these ever equal?

It seems to me that this doesn't fix the issue directly; it simply ensures that if there are already items in the tree, the cursor is never changed at all.

abought:
The preferences panel text change looks ok, but on initial read, the cursor fix may actually introduce a regression (it seems that if the client has fetched a list of nodes, it will never try to re-render the widget with an updated list).

We may need to focus on error handling in the AM, in which case we may want to remove the cursor stuff and move the remainder of the PR by itself. Can discuss briefly in case I'm misreading the logic here.

osfoffline/views/preferences.py
ticklemepierce: This is what prevents the cursor from being permanently a watch.  I talked to @chrisseto about it and this is what he helped me come up with.

abought:
Changes look good; works for me. @chennan47 , can you see if the loading cursor issue is also fixed on windows?

PR #60, opened 2015-11-13
Fix for two sync issues

himanshuo:
I am going through all the various sync types and making sure they are working. I have made subtasks on jira for these. The overarching task is (https://openscience.atlassian.net/browse/OSF-5127).  As I find the errors, I fix them.

The changes I have made so far are: 

1) using '/' for path instead of the '{node_id}:{provider}' 
2) By default, the category for a node is made to be a component in modify_local_node


abought:
This ticket is marked WIP- is this something we would like to include in next week's release?

himanshuo:
This exploratory ticket sought to find smaller issues. The two that I found are fixed by the current changes. The remaining related to moving nested files/folders. There is a ticket relating to that on jira now which I've been looking into (https://openscience.atlassian.net/browse/OSF-5190) + discussing resolutions for.

To answer your question, I think it would be good to include the few very small changes in this PR. Given that the other ticket is bigger and warrants more discussion, I think that should be handled separately.

abought:
Thanks. I try not to move forward WIP prs without asking; I'll change the title and take a look sometime today.

(Possibly after grocery shopping)

abought:
Notes for testers:

- As far as I can tell, item 1 refers to this ticket: https://openscience.atlassian.net/browse/OSF-5131 . The test case would seem to be moving a file from a subfolder (within a given component/project) into a parent folder.

abought:
I am still getting the database integrity error ([#OSF-5129]). We'll hold this PR until this can be explored further.

Feel free to comment if my testing notes above are incorrect; I'm trying to add specific actionable items so that QA can verify that bugs have been fixed.

abought:
I am still getting the `IntegrityError` described in https://openscience.atlassian.net/browse/OSF-5129 .

Specifically, I get this error when renaming a component (of type hypothesis) underneath a single synced top-level project. I get the error when I sync after renaming the component on the OSF.

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -61,8 +61,8 @@ def on_moved(self, event):
                     # This means it was put into a place on the hierarchy being watched but otherwise not attached to a
                     # node, so it needs to be added just like a new event rather than as a move.
 
-                    new_event = DirCreatedEvent(event.dest_path) if event.is_directory else FileCreatedEvent(event.dest_path)
-                    yield from self._create_file_or_folder(new_event, src_path=dest_path)
+                    # new_event = DirCreatedEvent(event.dest_path) if event.is_directory else FileCreatedEvent(event.dest_path)
+                    # yield from self._create_file_or_folder(new_event, src_path=dest_path)
```
abought: This change does not appear to reference either 5129 or 5131. What ticket is covered by these changes?

himanshuo: It was one of the things that broke in 5129. After the node category issue is fixed, this code ends up running and it creates a duplicate of the file/folder. However, I am NOT currently getting an error right now with that code. It is being run, but is not breaking anything. Thus, I have uncommented it.

PR #63, opened 2015-11-13
Handle problems when trying to get file updates from OSF

brianjgeiger:
Purpose
-----------
When bad things happen to good files on the OSF, the entire thread handling file download dies. Same going the other way. This handles the problems from the OSF side; another PR will handle the local to OSF path.

### DIFF
--- a/osfoffline/polling_osf_manager/polling_events.py
+++ b/osfoffline/polling_osf_manager/polling_events.py
@@ -35,7 +35,12 @@ class CreateFolder(PollingEvent):
             return
         if not os.path.exists(folder_to_create.full_path):
             AlertHandler.info(folder_to_create.name, AlertHandler.DOWNLOAD)
-            os.makedirs(folder_to_create.full_path)
+            try:
+                os.makedirs(folder_to_create.full_path)
+            except Exception:
+                # TODO: Narrow down this exception and do client side warnings
+                logging.exception('Exception caught: Invalid target path for folder.')
+                return
 
 
 class CreateFile(PollingEvent):
--- a/osfoffline/polling_osf_manager/polling_events.py
+++ b/osfoffline/polling_osf_manager/polling_events.py
@@ -124,7 +124,12 @@ class UpdateFile(PollingEvent):
         if not isinstance(self.download_url, str):
             logging.error('Update file download_url is not a str.')
             return
-        updated_file_path = ProperPath(self.path, is_dir=False)
+        try:
+            updated_file_path = ProperPath(self.path, is_dir=False)
+        except Exception:
+            # TODO: Narrow down this exception and do client side warnings
+            logging.exception('Exception caught: Invalid target path for updated file.')
+            return
         AlertHandler.info(updated_file_path.name, AlertHandler.MODIFYING)
         yield from _download_file(updated_file_path, self.download_url, self.osf_query)
 
--- a/osfoffline/polling_osf_manager/polling_events.py
+++ b/osfoffline/polling_osf_manager/polling_events.py
@@ -39,7 +39,7 @@ class CreateFolder(PollingEvent):
                 os.makedirs(folder_to_create.full_path)
             except Exception:
                 # TODO: Narrow down this exception and do client side warnings
-                logging.exception('Exception caught: Invalid target path for folder.')
+                logging.exception('Exception caught: Problem making a directory.')
                 return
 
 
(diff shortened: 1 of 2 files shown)


osfoffline/polling_osf_manager/polling_events.py
```diff
@@ -21,117 +21,150 @@ def run(self):
...
-        if not os.path.exists(self.path.full_path):
-            AlertHandler.info(self.path.name, AlertHandler.DOWNLOAD)
-            os.makedirs(self.path.full_path)
+        try:
+            folder_to_create = ProperPath(self.path, is_dir=True)
+        except Exception:
+            # TODO: Narrow down this exception and do client side warnings
+            logging.exception('Exception caught: Invalid target path for folder.')
+            return
+        if not os.path.exists(folder_to_create.full_path):
```
abought: For the `CreateFolder` event, you can make the test more specific by using `os.path.isdir` instead of `os.path.exists`.

brianjgeiger: True. I was trying not to otherwise affect functionality with this, only ensure that events are handled, but I can make changes if you'd like.

osfoffline/polling_osf_manager/polling_events.py
```diff
@@ -21,117 +21,150 @@ def run(self):
...
+        self.path = path
 
     @asyncio.coroutine
     def run(self):
         # create local node folder on filesystem
-        if not os.path.exists(self.path.full_path):
-            AlertHandler.info(self.path.name, AlertHandler.DOWNLOAD)
-            os.makedirs(self.path.full_path)
+        try:
+            folder_to_create = ProperPath(self.path, is_dir=True)
```
abought: Proper path does not perform OS-specific checks. It is possible that the program could try to create a file or folder with special characters (like `< > &`), and this error handler would not catch those problems.

Some error handling may need to be added at the actual act of creating the file/folder (eg line 38). This should be tested on MS Windows by trying to sync an OSF project with an oddly named file. I can provide a preconfigured VM next time we're in airdrop range, if that would save you time.

osfoffline/polling_osf_manager/polling_events.py
```diff
@@ -21,117 +21,150 @@ def run(self):
...
-        AlertHandler.info(self.path.name, AlertHandler.DOWNLOAD)
-        yield from _download_file(self.path, self.download_url, self.osf_query)
+        try:
+            new_file_path = ProperPath(self.path, is_dir=False)
+        except Exception:
+            # TODO: Narrow down this exception and do client side warnings
+            logging.exception('Exception caught: Invalid target path for new file.')
+            return
+        AlertHandler.info(new_file_path.name, AlertHandler.DOWNLOAD)
+        yield from _download_file(new_file_path, self.download_url, self.osf_query)
```
abought: As far as I can tell, Download file raises does handle OS errors (if it wouldn't be allowed to create the file locally). This can be tested by trying to sync a project anew from the OSF with a bad filename.

osfoffline/polling_osf_manager/polling_events.py
```diff
@@ -21,117 +21,150 @@ def run(self):
...
+            # TODO: Narrow down this exception and do client side warnings
+            logging.exception('Exception caught: Invalid source path for deleted folder.')
+            return
         # this works on systems that use file descriptors.
         # thus, linux, mac are supported.
         # todo: is windows supported??
         if shutil.rmtree.avoids_symlink_attacks:
-            AlertHandler.info(self.path.name, AlertHandler.DELETING)
+            AlertHandler.info(folder_to_delete.name, AlertHandler.DELETING)
             shutil.rmtree(
```
abought: This is where the error handling would most need to be added- the actual operation on the file.

abought:
I made a few notes in specific examples, but as a general rule, I think we'll need to focus a bit more error handling on the actual file operation, rather than just the (rather thin abstraction) that is `ProperPath`. This should be checked throughout the file. Some methods that use helper functions may be covered; other methods will not be.

We can discuss more if needed.

osfoffline/polling_osf_manager/polling_events.py
chrisseto: `os.makedirs(self.path.full_path, exist_ok=True)`

abought: > os.makedirs(self.path.full_path, exist_ok=True)

A new python3 flag? Nice.

osfoffline/polling_osf_manager/polling_events.py
brianjgeiger: Okay, I've added exception handling around line 38.

osfoffline/polling_osf_manager/polling_events.py
brianjgeiger: Everything in _download_file() is either a simple if statement or wrapped in a try-except without re-raising, so I think it should be okay.

osfoffline/polling_osf_manager/polling_events.py
brianjgeiger: Am I misunderstanding how `onerror=lambda a, b, c: logging.warning('local node not deleted because it does not exist.')` works? I can certainly wrap it, but I thought it would be redundant. That being said, I didn't do a lot of research into rmtree.

abought: Actually, I missed that line. A quick read suggests it does indeed work as described. :)

brianjgeiger:
If we hate these recent changes, we'll need to make a new ticket for fixing the problem Jinlu had and ditch the last two commits.

abought:
Ok to merge. It gives some error notifications, but we will need a follow up PR to handle an issue with bad filenames on windows.

PR #64, opened 2015-11-14
Exception handling for bad local files

mfraezz:
Purpose
=======
Prevent application from crashing when it could just ignore a file.

Changes
=======
Alerts the user that something went wrong whenever sync fails for file operations, and logs the exception.

Similar to https://github.com/CenterForOpenScience/OSF-Offline/pull/63

[#OSF-5194]

### DIFF
--- a/osfoffline/filesystem_manager/osf_event_handler.py
+++ b/osfoffline/filesystem_manager/osf_event_handler.py
@@ -185,7 +185,7 @@ class OSFEventHandler(FileSystemEventHandler):
             src_path = ProperPath(event.src_path, event.is_directory)
         except Exception:
             logging.exception('Exception caught: Invalid path')
-            AlterHandler.warn('invalid path specified. {} will not be synced.'.format(event.src_path.split('/')[-1]))
+            AlertHandler.warn('invalid path specified. {} will not be synced.'.format(event.src_path.split('/')[-1]))
         else:
             # create new model
             if self._already_exists(src_path):
@@ -209,7 +209,7 @@ class OSFEventHandler(FileSystemEventHandler):
             src_path = ProperPath(event.src_path, event.is_directory)
         except Exception:
             logging.exception('Exception caught: Invalid path')
-            AlterHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
+            AlertHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
         else:
             # get item
             try:
@@ -242,7 +242,7 @@ class OSFEventHandler(FileSystemEventHandler):
             src_path = ProperPath(event.src_path, event.is_directory)
         except Exception:
             logging.exception('Exception caught: Invalid path')
-            AlterHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
+            AlertHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
         else:
             if not self._already_exists(src_path):
                 return
--- a/osfoffline/filesystem_manager/osf_event_handler.py
+++ b/osfoffline/filesystem_manager/osf_event_handler.py
@@ -46,71 +46,94 @@ class OSFEventHandler(FileSystemEventHandler):
         :type event:
             :class:`DirMovedEvent` or :class:`FileMovedEvent`
         """
+        try:
+            src_path = ProperPath(event.src_path, event.is_directory)
+            dest_path = ProperPath(event.dest_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path specified')
+            AlertHandler.warn('Error moving {}. {} will not be synced.'.format('folder' if event.is_directory else 'file', src_path.split('/')[-1]))
+        else:
+            # determine and get what moved
+            if not self._already_exists(src_path):
+                try:
+                    self._get_parent_item_from_path(src_path)
+                except ItemNotInDB:
+                    # This means it was put into a place on the hierarchy being watched but otherwise not attached to a
+                    # node, so it needs to be added just like a new event rather than as a move.
 
-        src_path = ProperPath(event.src_path, event.is_directory)
-        dest_path = ProperPath(event.dest_path, event.is_directory)
-
-        # determine and get what moved
-        if not self._already_exists(src_path):
-            try:
-                self._get_parent_item_from_path(src_path)
-            except ItemNotInDB:
-                # This means it was put into a place on the hierarchy being watched but otherwise not attached to a
-                # node, so it needs to be added just like a new event rather than as a move.
-
-                new_event = DirCreatedEvent(event.dest_path) if event.is_directory else FileCreatedEvent(event.dest_path)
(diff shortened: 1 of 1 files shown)


mfraezz:
Testing note: easiest way to reproduce an issue that this is intended to catch is to include `..` in a filename (at least until OSF-5171 is fixed).

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -153,13 +181,17 @@ def on_created(self, event):
...
-        src_path = ProperPath(event.src_path, event.is_directory)
-
-        # create new model
-        if self._already_exists(src_path):
-            return
+        try:
+            src_path = ProperPath(event.src_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path')
+            AlterHandler.warn('invalid path specified. {} will not be synced.'.format(event.src_path.split('/')[-1]))
```
abought: Typo: `AlertHandler`

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -173,21 +205,29 @@ def on_modified(self, event):
...
         try:
-            item = self._get_item_by_path(src_path)
-        except ItemNotInDB:
-            # todo: create file folder
-            logging.warning('file {} was modified but not already in db. create it in db.'.format(src_path))
-            return  # todo: remove this once above is implemented
+            src_path = ProperPath(event.src_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path')
+            AlterHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
```
abought: Typo: `AlertHandler`

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -198,33 +238,40 @@ def on_deleted(self, event):
...
-                session.delete(item)
-                try:
-                    save(session)
-                except SQLAlchemyError as e:
-                    logging.exception('Error deleting node from database.')
+        try:
+            src_path = ProperPath(event.src_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path')
+            AlterHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
```
abought: Typo: `AlertHandler`

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -46,71 +46,94 @@ def on_moved(self, event):
...
-                # node, so it needs to be added just like a new event rather than as a move.
-
-                new_event = DirCreatedEvent(event.dest_path) if event.is_directory else FileCreatedEvent(event.dest_path)
-                yield from self._create_file_or_folder(new_event, src_path=dest_path)
+        try:
+            src_path = ProperPath(event.src_path, event.is_directory)
+            dest_path = ProperPath(event.dest_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path specified')
+            AlertHandler.warn('Error moving {}. {} will not be synced.'.format('folder' if event.is_directory else 'file', src_path.split('/')[-1]))
```
abought: `src_path` is a `ProperPath` object, and does not have a `split` method.

abought: Also, in general, you probably want `os.path.basename(somestr)` here (rather than splitting and extracting the last item yourself)

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -46,71 +46,94 @@ def on_moved(self, event):
...
-
-            # basically always osfstorage. this is just meant to be extendible in the future to other providers
-            item.provider = new_parent_item.provider if isinstance(new_parent_item, File) else File.DEFAULT_PROVIDER
-
-            # flags
-            item.locally_moved = True
-
-            save(session, item)
-            logging.info('moved from {} to {}'.format(src_path.full_path, dest_path.full_path))
+                logging.exception('Exception caught: Tried to move or rename item {}, but it could not be found in DB'.src_path.name)
```
abought: This should probably be `.format(...)`

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -46,71 +46,94 @@ def on_moved(self, event):
...
+                AlertHandler.warn('Could not find item to manipulate. {} will not be synced'.format(src_path.title))
```
abought: As far as I can tell, `ProperPath` objects do not have a title attribute.

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -173,21 +205,29 @@ def on_modified(self, event):
...
+                item = self._get_item_by_path(src_path)
+            except ItemNotInDB:
+                # todo: create file folder
+                logging.warning('file {} was modified but not already in db. create it in db.'.format(src_path))
+                return  # todo: remove this once above is implemented
 
-        # update hash
-        item.update_hash()
+            # update hash
+            item.update_hash()
```
abought: This function actually does open a function and try to update hash. It may fail if the file is for some reason inaccessible; may be worth catching an `OsError` here.

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -276,11 +321,14 @@ def _get_item_by_path(self, path):
...
-            return True
         try:
-            if ProperPath(event.dest_path, True).name == 'Components':
+            if ProperPath(event.src_path, True).name == 'Components':
                 return True
-            return False
-        except AttributeError:
-            return False
+            try:
+                if ProperPath(event.dest_path, True).name == 'Components':
```
abought: I think that these three lines can be consolidated to 
`return ProperPath(event.dest_path, True).name == 'Components'`

osfoffline/filesystem_manager/osf_event_handler.py
```diff
@@ -198,33 +238,40 @@ def on_deleted(self, event):
...
-                session.delete(item)
-                try:
-                    save(session)
-                except SQLAlchemyError as e:
-                    logging.exception('Error deleting node from database.')
+        try:
+            src_path = ProperPath(event.src_path, event.is_directory)
+        except Exception:
+            logging.exception('Exception caught: Invalid path')
+            AlertHandler.warn('invalid path specified. {} will not be synced'.format(event.src_path.split('/')[-1]))
```
abought: Any reason not to use `os.path.basename` here to get the last part of the path?

abought:
As far as I can tell, `ProperPath` only raises one type of exception- `InvalidPathError`. In that case we may want to stick to specific handling rather than bare `Exceptions` in our try/except blocks.

abought:
Pass finished.The two biggest requests: one new item of error handling might be helpful to add, and a few statements seem to have typos that would prevent the code from actually running. 

Otherwise error handling seems to cover the major cases. (we'll worry about making it look pretty post-alpha)

abought:
Changes look good, worked with the simple test case specified. We'll move it forward for QA to test in depth.

PR #70, opened 2015-11-16
[Fix]Allow multiple uploads

mfraezz:
Purpose
=======
Allow multiple uploads. You can only key one thing to `None`, and if a file is newly created locally, it won't have an `osf_id` until it's pushed to the OSF.

Changes
=======
Ensure all such files are added to the tuple list.

Side Effects
==========
None


abought:
Code looks good, demo passed.

PR #72, opened 2015-11-16
[hold][Feature] Errors for illegal characters/Microsoft's reserved words

ticklemepierce:
Purpose
-
Microsoft has a bunch of rules on file system naming (https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx).  This PR addresses the rules for illegal names and illegal symbols.  

Changes
-
Made it so that the Settings panel won't display a node if the node name contains an illegal symbol (as it wouldn't be able to be created anyways) and then in polling.py created checks before attempting to create the file/folder to see if it's name is valid.  The reason for this is to keep the same namespace available on Windows and Mac/Linux.

Side Effects
-
Currently no check is performed when renaming to see whether the new name is valid or not (not that we can stop this anyways - just provide an error) so if the user names something on Mac with an illegal name or symbol and then someone attempts to sync this file on Windows, it will fail on Windows (but will fail gracefully because of this PR).  Ideally, the user should be prompted that their name change may cause this to occur.  Other than that no potential side effects are known.

### DIFF
--- a/osfoffline/polling_osf_manager/polling.py
+++ b/osfoffline/polling_osf_manager/polling.py
@@ -424,33 +424,38 @@ class Poll(object):
 
         # NOTE: develop is not letting me download files. dont know why.
 
-        # create local file folder in db
-        file_type = File.FILE if isinstance(remote_file_folder, RemoteFile) else File.FOLDER
-        new_file_folder = File(
-            name=remote_file_folder.name,
-            type=file_type,
-            osf_id=remote_file_folder.id,
-            provider=remote_file_folder.provider,
-            osf_path=remote_file_folder.id,
-            user=self.user,
-            parent=local_parent_folder,
-            node=local_node
-        )
-        save(session, new_file_folder)
-
-        if file_type == File.FILE:
-            event = CreateFile(
-                path=new_file_folder.path,
-                download_url=remote_file_folder.download_url,
-                osf_query=self.osf_query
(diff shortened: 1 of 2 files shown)


abought:
To help QA verify this fix, can you describe what test cases you used when running this (on Windows) to validate the intended functionality?

osfoffline/polling_osf_manager/polling.py
```diff
@@ -371,25 +371,23 @@ def _check_file_folder(self,
         else:
             raise ValueError('in some weird state. figure it out.')
 
-        assert local_file_folder is not None
-        assert remote_file_folder is not None
```
abought: The assertions prevented this function from *ever* being used if the preconditions were not met, whereas nested things inside an if block means that a failure will pass silently.

I'm not a fan of assertions for general checks (like `try: raise AssertionError  / except AssertionError as e`, but in this case they provide a very strict sanity check. I'm not convinced that the if statements (which fail silently if the condition is not met) are a complete replacement.

abought:
Looking forward to reviewing this, but I'm going to make a call to hold this one out of the deployed version for tomorrow AM; it's already 9pm and I'm still in the office. (we need to actually deploy something today!)

We can take a closer look in the AM, because there is life after alpha! If bad filenames remain an issue this will be a priority to shift by the end of the day. Will keep you posted.

abought:

PR #74, opened 2015-11-17
[Fix] Remove limit on queue size

mfraezz:
Purpose
======
Bug found where only 16-17 files were downloaded, due to the queue decrementing task count when queue was at max size and one (finished) job was replaced with a new job.

Changes
=======
* `maxsize=0` means queue is unbounded
* `try/finally` to ensure job is marked as down even if an exception is raised.

Side Effects
=========
If more jobs are queued than your computer has memory for (assuming max 1kb per job), memory overflow is possible.

### DIFF
--- a/osfoffline/polling_osf_manager/polling.py
+++ b/osfoffline/polling_osf_manager/polling.py
@@ -82,13 +82,12 @@ class Poll(object):
         self._loop.call_later(5, self.start)
 
     def start(self):
-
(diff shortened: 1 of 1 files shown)


abought:
As discussed in person, looks reasonable.

I agree that removing the limit seems like it could be a problem, but this will have to do until a possible asyncio bug is fixed.

PR #75, opened 2015-11-17
Fix python3.4-5's queue issue

chrisseto:


abought:
Note for testers: this fixes a bug where large sync events would not get every file. The fix is essentially a polyfill based on more recent versions of python.

(example: 5 projects of 5 files each)

PR #76, opened 2015-11-17
Allow clarg to drop db

mfraezz:
start app with 
```bash
open /Applications/OSF-Offline.app --args --drop
```
to drop DB on start. `--args` tells `open` to pass `--drop` to `OSF-Offline.app`.

### DIFF
--- a/osfoffline/database_manager/db.py
+++ b/osfoffline/database_manager/db.py
@@ -1,3 +1,5 @@
+import contextlib
+
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker, scoped_session
 from osfoffline.database_manager.models import Base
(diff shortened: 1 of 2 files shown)


abought:
Note for testers: this is not a feature we intend for people to use everyday. It is intended as a last-resort way for a user to reset the application (when the state of their database becomes corrupted) without them having to dig through files that they may not even be able to find.

chennan47:
The only concern for this is that after deploy. It will be OSF-Offline.exe and based on different platform it will be running. This command will be different?

abought:
It'll be similar, probably just omitting the word "open" (and specifying .exe instead of .app)

abought:
Works at command line on mac, and clear from me, pending @chennan47 test on windows.

chennan47:
Verified it works on windows deployed exe version. But the key word is "start"

PR #77, opened 2015-11-17
Use ProperPath.full_path, not str.full_path

mfraezz:


abought:
Simple and straightforward; if it fixes an active bug that's a bonus.

PR #78, opened 2015-11-17
Fix issue with moving file to a parent folder

abought:
Refs [#OSF-5131]. This PR fixes an error that would be encountered when moving a file from a subfolder to a parent directory.

Forked from #60 per discussion with @himanshuo


chennan47:
This fix doesn't seem to work on either Mac or windows. The sync is not moving the files at all.

PR #82, opened 2015-11-17
[Feature]Redownload file if it is expected to be there

mfraezz:
Purpose
======
If a download fails, we should try to download it on subsequent syncs.

Changes
=======
A failed download will be retried on next sync

Side effects
=========
A file that was deleted when OSFO was not watching will be redownloaded (probably desired behavior)

Test Notes
---------------
Some fixes have gone in to limit the number of failed downloads, but it's still possible in the case of a failed connection, and possibly other cases. In any case, when a `File` object is successfully added to the database but not the filesystem, we should retry the download.

1) I un-fixed the bug from #75 for testing, but disconnecting wifi in the middle of a sync should work just as well. With several failed downloads, there should be log statements similar to `file {filename} not reuploaded on remote server because does not exist locally.` in the log file. Press `Sync Now` while using this code, and the listed `{filename}`s should be marked for download again.

2) Open `Settings` or close application to halt polling. Delete a file locally. When syncing restarts (when the application restarts), the deleted file should be downloaded again. Note: this does not affect files deleted **while** the application is watching

### DIFF
--- a/osfoffline/polling_osf_manager/polling.py
+++ b/osfoffline/polling_osf_manager/polling.py
@@ -585,7 +585,7 @@ class Poll(object):
             new_remote_file = yield from self.osf_query.upload_file(local_file)
         except FileNotFoundError:
             logger.warning(
-                'file not reuploaded on remote server because does not exist locally. it probably downloaded incorrectly. inside create_remote_file_folder')
+                'file {} not reuploaded on remote server because does not exist locally. it probably downloaded incorrectly.'.format(local_file.name))
             yield from self.update_local_file(local_file, remote_file)
             return remote_file
 
--- a/osfoffline/polling_osf_manager/polling.py
+++ b/osfoffline/polling_osf_manager/polling.py
@@ -6,6 +6,7 @@ import time
 
 import aiohttp
 import iso8601
+from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 import osfoffline.alerts as AlertHandler
@@ -585,7 +586,8 @@ class Poll(object):
             new_remote_file = yield from self.osf_query.upload_file(local_file)
         except FileNotFoundError:
             logger.warning(
-                'file not reuploaded on remote server because does not exist locally. inside create_remote_file_folder')
+                'file not reuploaded on remote server because does not exist locally. it probably downloaded incorrectly. inside create_remote_file_folder')
+            yield from self.update_local_file(local_file, remote_file)
             return remote_file
 
         return new_remote_file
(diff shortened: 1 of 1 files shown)


osfoffline/polling_osf_manager/polling.py
```diff
@@ -6,6 +6,7 @@
 
 import aiohttp
 import iso8601
+from sqlalchemy.exc import SQLAlchemyError
```
abought: This import is unused.

mfraezz: Started to make the `save`s safer, but those exceptions should probably be raised. I'll remove this

osfoffline/polling_osf_manager/polling.py
```diff
@@ -585,7 +585,8 @@ def update_remote_file(self, local_file, remote_file):
             new_remote_file = yield from self.osf_query.upload_file(local_file)
         except FileNotFoundError:
             logger.warning(
-                'file not reuploaded on remote server because does not exist locally. inside create_remote_file_folder')
+                'file not reuploaded on remote server because does not exist locally. it probably downloaded incorrectly. inside create_remote_file_folder')
```
abought: What does "reuploaded on remote" mean?

abought: Please add the name of the file to the warning if possible- the logs are more useful if we know which file is having the problem. (fileobj.name or fileobj.path)

mfraezz: It thinks that the local version of the file is ahead of the remote because it didn't update the last_modified because the download never completed.

Will add fillename

abought:
Note to self for review after lunch:

This line fires when the following conditions are met:
1. The local and remote files both exist
2. The local file timestamp is newer than the remote file (according to the database) [[ref](https://github.com/abought/OSF-Offline/blob/develop/osfoffline/polling_osf_manager/polling.py#L550)]
3. In the process of attempting to upload the local file to the server, it is discovered that the local file doesn't actually exist.
4. Instead of updating the remote file, the purpose of the function flips on its head and instead tries to update/fetch the local file

Where in this chain does the database entry describing the local file change? It seems to me that if the server were updated with a newer version, this would lead to a situation where the database no longer reflected the state of content fetched from the OSF.

Example:
- User fails to download file on first try. Database stores modified date of x, which was the change time of the first download attempt.

- Another user uploads a new version on the OSF (modified time y)

- User A tries to sync according to the rules of this ticket, DB entry doesn't appear to be updated, and their database now thinks that version Y was actually last updated at time X.

Will discuss after lunch in case I'm missing an aspect here.

mfraezz:
In that case, it should just try to fetch the remote (also with `update_local_file`) and miss this branch of logic entirely, because https://github.com/mfraezz/OSF-Offline/blob/feature/retry-download-if-fail/osfoffline/polling_osf_manager/polling.py#L532-L536 would detect that the remote file was newer.

PR #83, opened 2015-11-17
[Feature] Drop first user on login of second

mfraezz:
Purpose
=======
If a second user tries to log in, drop all db models relating to first, and allow login

Changes
=======
* Support for above
* Helper fucntion that drops db without causing the next `save` to raise errors

Side Effects
==========
None

### DIFF
--- a/osfoffline/database_manager/__init__.py
+++ b/osfoffline/database_manager/__init__.py
@@ -1,1 +1,3 @@
-__author__ = 'himanshu'
+from osfoffline.database_manager.models import User, Node, File
+
+CORE_OSFO_MODELS = [User, Node, File]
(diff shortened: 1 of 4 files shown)


chennan47:
A small problem needs to be fixed for this. Whenever you log in as second user, the project showed up on the syncing list will be updated a few secs after you select osf tab on settings which means the second user has a change to take a glance at what first user's project are. And if you don't make any selection for projects to sync and close the settings window directly. It will still say if you want to save the changes for the project to sync you have made. @mfraezz

mfraezz:
@chennan47 Both of those issues have been addressed. Good catch

chennan47:
@abought first pass done and functional review done. Looks good on Mac and Windows. Feel free to merge it if the code is cool with you.

PR #84, opened 2015-11-17
[Fix] File Notifications

mfraezz:
Delete file wasn't giving user a notification


osfoffline/polling_osf_manager/polling_events.py
```diff
@@ -179,6 +179,7 @@ def __init__(self, path):
     @asyncio.coroutine
     def run(self):
         file_to_delete = ProperPath(self.path, is_dir=False)
+        AlertHandler.info(file_to_delete.name, AlertHandler.DELETING)
```
abought: No side effects; ok here.

PR #92, opened 2015-11-18
Remove symlink attack prevention

mfraezz:
`shutil` automatically uses attack prevention if it can (OSX)
We want it to still remove stuff if it can't (Win32)


abought:
Code looks good. Ping me when it runs.

abought:
Future note: we will need to revisit this removal in the future if we target more operating systems.

For example, an OS that allows symlink attacks (but does not support the safe method options to mitigate).

abought:
Also, at the end of the day we can worry about bringing develop up to date with new master commits.

PR #97, opened 2015-11-19
[WIP][Feature] Init sync

chrisseto:


osfoffline/views/preferences.py
```diff
@@ -219,7 +218,7 @@ def fetch(self):
                 user_nodes = []
                 url = api_url_for(USERS, related_type=NODES, user_id=user.osf_id)
                 headers = {'Authorization': 'Bearer {}'.format(user.oauth_token)}
-                resp = requests.get(url, headers=headers).json()
+                resp = requests.get(url, headers=headers, params={'page[size]': 500}).json()
```
abought: :octopus: 

Larger page size in queries = whee!

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
     DirCreatedEvent,
 )
 from watchdog.observers import Observer
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
```
abought: Various imports in this section are no longer used, most notably deletion events. Is that an intentional design change?

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
...
+        for (name, is_dir), (local, remote) in resources.items():
+            if is_dir:
+                directories.append((
+                    getattr(local, 'fullpath', None),
+                    getattr(remote, 'osf_id', None)
+                ))
+            event = self.infer_event(local, remote)
+            if event is not None:
+                events.append(event)
+                # DO SOMETHING HERE
```
abought: This class doesn't seem to have a queue at all; are the comments on these lines still relevant in any way?

osfoffline/application/background.py
```diff
@@ -55,6 +55,8 @@ def run(self):
         self.user = self.get_current_user()
         self.osf_folder = self.user.osf_local_folder_path
 
+        LocalDBSync(self.observer, self.user, self.loop).sync_db()
```
abought: (comment duplicated from #55)

This is probably not the best place to call `LocalDBSync`. We want to use initial sync only when the program first starts, but putting here means it may also be called when we click sync now, when the prefs panel is opened/closed, etc. This may create a situation where the "initial" sync (this class) generates events that do not match "active sync" (watchdog file monitoring, eg move and deletes).

We will need to test carefully and revisit best place for this code once the program restructuring is done. After merge, it may make sense to turn off `LocalDBSync` for now. (keep the refactored logic, but don't use it yet)

This line appears in BackgroundWorker.run(...), which is called when the thread is started. Eg:
1. OsfApp.start (which is our intended goal)
2. OsfApp.resume (prefs panel close, syc now, etc)

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
...
+        for local in local_list:
+            ret.setdefault((local.name, local.is_dir), [None, None])[0] = local
+
+        for remote in remote_list:
+            ret.setdefault((remote.name, remote.is_dir), [None, None])[1] = remote
+
+        return ret
+
+    @asyncio.coroutine
+    def crawl_node(self, node, local_path='/', remote_path='/'):
```
abought: Question to be sure I'm understanding this right:

This method only seems to be used in blocking ways (`run_until_complete`, `yield from`, etc), and by design it seems that we would only ever want it to run by itself. What is the advantage of making it a coroutine?

abought: Disregard above- coroutine appears to be a requirement to call `list_remote_dir` in particular, which makes a network request.

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
...
+    def sync_db(self):
+        logger.info('Beginning initial sync')
+        events, nodes = [], [
+            node for node in
+            self.user.top_level_nodes
+            if node.osf_id in self.user.guid_for_top_level_nodes_to_sync
+        ]
+        for node in nodes:
+            logger.info('Resyncing node {}'.format(node))
+            events.extend(self.loop.run_until_complete(self.crawl_node(node)))
```
abought: Note to self: check this during functional testing step.

Watchdog events appear to be just data objects- although these objects are created (in `infer_event`), I'm not clear on whether the events are actually being fired in a way that lets `osf_event_handler` pick up on them.

In the old version we [added them](https://github.com/CenterForOpenScience/OSF-Offline/pull/97/files#diff-ad501704142723e44bdb2e2d5267671eL171) to an event queue on the observer, but this appears to no longer be the case, and the value of `events` does not appear to be used (or returned) by the `sync_db` function.

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
...
-        if local and db and self._get_proper_path(local) != self._get_proper_path(db):
-            raise IncorrectLocalDBMatch
-        if local and not isinstance(local, ProperPath):
-            raise InvalidItemType
+        if remote_folder != '/':
+            return (yield from self.osf_query.get_child_files(remote_folder))
 
-        local_children = {}
-        db_children = {}
+        if not hasattr(self, '__nodes'):
```
abought: Note to self:  define `self._nodes = None` in the `__init__` method and remove the `hasattr` check.

Otherwise, since we don't reuse the `'LocalDBSync` instance, caching in this way seems fine.

osfoffline/filesystem_manager/sync_local_filesystem_and_db.py
```diff
@@ -11,209 +13,185 @@
...
-                event = FileModifiedEvent(self._get_proper_path(local).full_path)  # create changed event
-                # folder modified event cannot happen. It will be a create and delete event.
-        elif local is None:
-            db_path = self._get_proper_path(db).full_path
-            if isinstance(db, File) and db.is_file:
-                event = FileDeletedEvent(db_path)  # delete event for file
+                return None
+        except (NoResultFound, StopIteration):
+            if local_file and remote_file:
+                # pls to implement
```
abought: Note to self: address this comment

abought:
I might just be missing this, but `crawl_node` seems to match local folders (first level) with remote nodes (using an endpoint that only gives top level nodes). 

Where does this recurse through child nodes, to be sure that sub-components are also being synced? (children are represented as `Node.child_nodes` in the database, or `osf_query.get_child_nodes` / `RemoteNode.child_nodes_url` in the api client)

Hopefully I'm missing something obvious. :)

abought:
Review paused for me for now- need to do some functional testing next. A few questions and comments, and areas where functional testing may be needed to ensure compatibility with original design.

Since Chris likely won't have time to work on this: if we do pick it up (after merging in Michael's changes), we should turn off ` LocalDBSync` while making modifications to sync logic as needed.

PR #104, opened 2015-11-20
[WIP] Make move events be top down instead of bottom up OSF-5190

himanshuo:
See osf-5190 on jira 

Problem Statement:

OSFEventHandler._get_parent_item_from_path(dest_path) works by simply stripping the rightmost item in the dest_path and then searching the database for an item with that new path. This fails when you are moving an entire directory with nested files because watchdog sends events for each file in the nested hierarchy. For example, say we have the folder structure:
A
-----C
--------myfile.txt
B 
and we move C to be under B. Watchdog will give us the events the following order:
1) Moved file: A/C/myfile.txt to B/C/myfile.txt
2) Moved directory: A/C to B/C

We would try and work with the file moved event first in the code, however, the parent for the destination (B/C) does not exist and so we end up raising an error.
What would be preferred is that watchdog only send us the event for moving the directory A/C to B/C because technically that is all that we were trying to do. The internal files just happened to be there, but we do not really care for them. If not that, then it could also work if watchdog gives us the events in reverse order - the folder and then the internal file. This would allow us to ensure that the folder is moved in the database before moving the file.

### DIFF
--- a/osfoffline/application/background.py
+++ b/osfoffline/application/background.py
@@ -3,8 +3,11 @@ import asyncio
 import logging
 import threading
 
-from watchdog.observers import Observer
-
+from watchdog.observers.api import (
+    BaseObserver,
+    DEFAULT_OBSERVER_TIMEOUT
+)
+from osfoffline.filesystem_manager.top_down_move_event_emitter import TopDownMovePollingEmitter
 
 from osfoffline.database_manager import models
 from osfoffline.database_manager.db import session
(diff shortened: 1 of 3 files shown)


abought:
As discussed in person, how does this fit into the proposed refactor for consolidating events?  See:

https://github.com/CenterForOpenScience/OSF-Offline/blob/develop/osfoffline/sync/watchdog.py#L10

PR #105, opened 2015-11-30
General cleanup

abought:
Tracking branch; not ready to merge.

## Purpose
Before we begin to implement tasks, performed some small cleanup to make codebase work better with shared settings file, and pass flake8 checks.

## Summary of changes
- Changed `tui.py` to use shared settings files and respect user config. Also some minor cleanup of old deprecated settings.
- Changed all kwargs to be keyword-only (to encourage good coding practices later)
- Sort imports and requirements files


osfoffline/utils/authentication.py
```diff
@@ -118,7 +123,7 @@ def log_in(self, user=None, username=None, password=None):
         user.logged_in = True
         try:
             save(session, user)
-        except SQLAlchemyError as e:
+        except SQLAlchemyError as e:  # noqa
```
chennan47: Not clear what this comment means

abought: "noqa" tells flake (our syntax checker) to skip this line. Otherwise we get an error ("variable e is defined but never used").

PR #109, opened 2015-12-07
[review] Log errors to sentry

abought:
## Purpose
Add Sentry to logging module (must specify DSN in `local.py`- don't commit to github). This will ensure that exceptions are reported to a central server.

Refs https://openscience.atlassian.net/browse/OFF-47

## Summary of changes
- Add Raven as dependency (pin to a version compatible with our Sentry server- newer versions will not work)
- Move `LOGGING_CONFIG` to `settings/__init__.py`, so as to respect user settings overrides (for level or sentry dsn)
- Use raven's integrated logging handler
- When user logs in, add user data to all sentry messages across threads
- Restore `start_for_tests` as an invoke task, to aid in development

## Design/ deployment notes
Used Sentry's logging integration, so that errors will be tracked consistently whether they happen while connected to the internet or not.

The Sentry DSN is available on request. It should be specified in the user's `local.py` file, but not committed to Github. (local.py will need to be modified when preparing a custom build for deployment)

## To do
- [x] Investigate whether sensitive parameters (eg user credentials) are included in error messages that may need to be scrubbed
- [x] Add specific extra information (eg username) to logs if possible, to help track support incidents. (It does automatically record server/machine name, which may or may not be descriptive)
- [x] Test log format in depth; will use a standalone Sentry instance until issue with detail view is resolved
- [x] Ensure no duplicate events are logged
- (future) Add filters for other types of sensitive data


abought:
Testing notes:

Because our hosted instance gives 502 errors when trying to view the detail page on a sentry error, I ran this code using Sentry's hosted service. 

I checked the following behaviors, and examined error traces for each. Passwords are automatically filtered. (will evaluate tokens in a followup PR)

Tried the following to ensure behavior was as expected:
- Confirmed program runs (with sentry warning) when no DSN is specified, and logs to server correctly when a DSN is provided
- Fired an exception before login (no user data included), on the UI thread; ensured it was logged
- Fired exception after login (user data included),  on the UI thread;  ensured it was logged (including username)
- Fired exception after login, on the background worker thread;  ensured it was logged (including username)

abought:
Question for reviewers: do we plan to add a configuration option (so users can choose not to report errors to sentry)?

If so, we may want to add a placeholder field to the user model. That would remove the need to write a migration later.

abought:
Tagging @mattclark for visibility: fixing sentry detail page 502 errors would be very helpful for a successful deployment of this PR.

osfoffline/settings/__init__.py
```diff
@@ -16,4 +19,44 @@
...
+        'handlers': ['console', 'logfile']
+    }
+}
+
+# Set up basic (console and file) logging
 logging.config.dictConfig(LOGGING_CONFIG)
+
+# Add Sentry logging separately, so that we can access the client and modify context variables later
+# This allows us to send additional data to Sentry (like username, when the user is logged in)
+# TODO: We should allow the user to choose whether they log to sentry
```
icereval: Agreed, logged as v1 blocker [#OFF-74].

abought: > Agreed, logged as v1 blocker [#OFF-74].

Would you like me to create a JIRA ticket separate from [#OFF-74] to track followon improvement work?

abought: > Would you like me to create a JIRA ticket separate from [#OFF-74] to track followon improvement work?

Nevermind; got my tickets mixed up. Very sorry.

PR #110, opened 2015-12-10
Add support for two-factor authentication

abought:
## Purpose
Respect two-factor settings for users with 2FA login enabled. Allow these users to use OSF offline client.

## References
Ticket: https://openscience.atlassian.net/browse/OFF-46

## Summary of changes
- Add UI input box to request 2FA one-time password as needed.
- Some minor internal cleanup and changes to method signatures
- Fix regression bug, where clicking logout did not actually log the user out. (db changes weren't being saved)

## Design and testing notes
The client makes an entire separate request to determine whether or not a 2-factor code is necessary. The overall UI flow (2FA request on next page) is similar to how the OSF website presents the form field.

Test cases:
- As a user without 2FA enabled
  - Try to log in with the wrong password (login fails)
  - Try to log in with the right password (login succeeds)
- As a user with 2FA enabled
  - Try to log in with the wrong password (login fails)
  - Try to log in with the right password- presented with a prompt to enter 2FA code
    - Enter the wrong 2FA code, get prompted again
    - Cancel the dialog box, get taken back to the login screen
    - Enter the right 2FA code, successfully log in
  - Close the client and reopen. Should not be prompted for a 2FA code, because user is now using personal access token.
- Log out of the client, and restart the program. Should be taken to a login screen. (fixed regression unrelated to other features)


chennan47:
Code review done. Looks good.

PR #111, opened 2015-12-14
Exclude registrations from API request

abought:
## Purpose
Fix an issue where the application would crash when viewing list of nodes to sync, if a user owned any registrations.

Modified the API call to exclude any registrations from the list of user nodes.

## Scope
Currently, the API attempts to act on data that may not have been returned from the server. Preferences/py does not currently handle exceptions correctly; that should be fixed separately as part of a broader error handling task.

For now, just changed logging so that we capture the traceback correctly.


chennan47:
Review done. Looks good.

PR #114, opened 2015-12-16
[review] Basic implementation of consolidated notifications

abought:
Refs https://openscience.atlassian.net/browse/OFF-48

## Purpose
Consolidate close-spaced notifications together. Give the user one message, instead of 10.

As is, this doesn't apply any special logic to the grouping. It simply grabs all notifications that occur in a certain span of time. There is an assumption that anything in `notification_queue` must be a file or folder sync operation. CRUD operations are grouped together in a single counter. See notification center screenshot below for example.

screen shot 2015-12-16 at 4 48 40 pm

## Summary of changes
- When a notification fires, wait several seconds and gather all notifications together in a burst. Ensures that the user never has to wait too long for feedback on progress by using a straight timer; may implement debounce option later.
   - INFO messages can be grouped, but messages of a higher severity will always be displayed individually.
- Settings cleanup
   - Remove unused settings
   - Tweak format of log messages
   - Change ALERT_TIME to ALERT_DURATION, and unit to seconds (more consistent with python APIs)

## Test cases
Sync a large project. Try several different ALERT_DURATION values and see the size of the groups change.

## TODO
- [x] Fix straggler bug. Traced to here: https://github.com/CenterForOpenScience/OSF-Offline/blob/develop/osfoffline/gui/qt/tray.py#L194 (one event is being pulled from the queue, out of sequence, before ready)

### DIFF
--- a/osfoffline/gui/qt/tray.py
+++ b/osfoffline/gui/qt/tray.py
@@ -175,8 +175,8 @@ class OSFOfflineQT(QSystemTrayIcon):
                     self._show_notifications(notification_list)
                 else:
                     self.showMessage(
-                        'Updating multiple',
-                        'Updating {} files and folders'.format(len(notification_list)),
+                        'Updated multiple',
+                        'Updated {} files and folders'.format(len(notification_list)),
                         QSystemTrayIcon.NoIcon,
                         msecs=settings.ALERT_DURATION / 1000.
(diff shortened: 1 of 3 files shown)


icereval:
certainly a good first step to not spam the user, looks good!

PR #115, opened 2015-12-17
[review] Cleanup and pruning

abought:
## Purpose
Various simple code cleanup and removal of unused files. (housekeeping during readthrough)

## Summary of changes
- Add an invoke task for flake8 syntax checking
  - Resolve outstanding flake errors
  - Use settings consistent with other OSF projects
- Delete unused or outdated files
- DRY function definitions


tui.py
```diff
@@ -1,232 +0,0 @@
-import collections
```
icereval: why would you remove the nearly complete v1 osf-offline console app?

abought: In a conversation yesterday, I got the impression that it was safe to remove insofar as you had made the decision to go with QT for development instead.

Separately, it's a second frontend codebase to maintain as program specifications evolve. Just seemed like more overhead to be lugging around, but if it's actively being used and necessary, I can add it back.

icereval: I guess I can move on knowing tui.py went out at its peak and being forever preserved in glorious git history, long live tui.py!!!

abought: Tell you what: if we make it an entire month into the beta without anyone egging our office in retaliation for bugs, I'll order you a custom commemorative t-shirt.

(the "t" stands for "tui")

PR #116, opened 2015-12-17
Add various error handling

abought:
Refs
https://openscience.atlassian.net/browse/OFF-72
https://openscience.atlassian.net/browse/OFF-65
https://openscience.atlassian.net/browse/OFF-76 

## Purpose
Add various forms of error handling throughout the application.

## Limitations
This PR focuses on very broad error handling- it will prevent the program from crashing inexplicably, but the user may not always know why something is failing. 

Error messages are crude, and exceptions are caught at a level somewhat removed from the point of failure.

## Test cases that should be fixed by this PR
- [x] Timeout error during remote sync operation hangs the background thread and prevents later sync attempts (to test: activate "100% loss" profile in network link conditioner, try to sync, then subsequently try to sync when connection is restored)
- [x] Set user data folder to a destination where the user doesn't have permission (such as root directory or network volume) causes program to immediately crash- user is now warned about permissions, and given chance to select new folder
- [x] Clicking cancel on the "Choose sync folder" dialog should not result in using a default (mystery) folder whose location is not communicated to the user.
- [x] User selects a sync folder that they no longer have write access to (eg a removable hard drive that was unplugged) and application hangs- this will now- however briefly- notify the user that an error occurred during the sync process
- [x] API error while loading preferences node list should not cause the application to crash (recent example: #111 - can be tested by using Network link conditioner to disable internet connection, then visiting the settings tab of the preferences panel. User should now receive a warning message). 
- [x] A critical failure in syncing one project will prevent other projects from syncing at all. (example: delete a project, then add another project. If the deleted project is synced first, the second project would not be attempted) [refs OFF-76](https://openscience.atlassian.net/browse/OFF-76)

## Summary of other changes
- Some small cleanup to log messages. 
- Log in button on preferences panel should now function as expected. ([#OFF-65])

## Review notes
This does add calls to `Notification` in a few new places throughout the code; there may be room to improve separation of concerns.

The fix in 95e862d is provisional; it merits a further look by someone more familiar with the auditor code.

## To Do
- [ ] Improve user-facing error message for file-operation job failure; include filename where possible.


osfoffline/tasks/operations.py
```diff
@@ -121,7 +121,6 @@ class LocalCreateFile(BaseOperation):
     """Download an individual file from the OSF into a folder that already exists"""
 
     def _run(self):
-        logger.info('LocalCreateFile: {}'.format(self.remote))
```
samchrisinger: I've found this logging to be very useful during development.

abought: Agreed! We were actually logging this info more than once, and I was trying to consolidate it- putting the message in the execution of the job queue instead of in the job method itself. Let me know if you'd prefer it be here instead?

https://github.com/CenterForOpenScience/OSF-Offline/pull/116/files#diff-ee86bf30e5e21f98e3c70a670a827861R32

PR #117, opened 2015-12-18
Add basic deployment notes

abought:
Refs https://openscience.atlassian.net/browse/OFF-13

Covers building an installer on Windows and Mac from scratch, including software requirements to prepare the build and dev environment.

Also update requirements.txt to remove unused requirements, and avoid the need to manually edit pyinstaller source code on Windows.


docs/deployment/DeploymentGuide.md
```diff
@@ -0,0 +1,89 @@
...
+  recommended for Windows operating systems, for best compatibility with precompiled Windows PyQT binaries.
+  - On Windows, be sure to select a custom install, and choose the option to add Python (and script folders)
+  to the system PATH automatically.
+- `pip`
+  - Should already come with python. Make sure it can be run via the command line.
+- `git`
+  - On Windows, the [GitHub desktop client](https://desktop.github.com/) provides a viable installation. (default
+  data folder is `My Documents\GitHub`)
+- [qt5.5](http://www.qt.io/download-open-source/)
+  - Can de-select QT 5.4 from the installer to save disk space; only need 5.5. Other options left as-is.
```
chennan47: Suggest to de-select 5.4 in case the wrong configure will cause the usage of 5.4 instead of 5.5 which has a problem with system tray icon disappear sometime.

PR #119, opened 2015-12-29
[WIP] Add support for syncing components

samchrisinger:
# Depends on:
https://github.com/CenterForOpenScience/OSF-Offline/pull/120/files


osfoffline/client/osf.py
```diff
@@ -59,9 +64,38 @@ def load(cls, request_session, *args):
...
+        return cls.from_data(request_session, data['data'])
+
+    def _fetch_related(self, relationship, data):
+        items = data['data']
+        for item in items:
+            yield item
+        if data['links'].get('next'):
+            for item in self.fetch_related(
+                    self.request_session,
+                    relationship
```
chrisseto: This looks like it will go into an infinite loop as no page is provided to `fetch_related`

osfoffline/client/osf.py
```diff
@@ -88,10 +122,19 @@ def __init__(self, request_session, data):
         super().__init__(request_session, data)
         self.date_created = iso8601.parse_date(self.date_created)
         self.date_modified = iso8601.parse_date(self.date_modified)
+        self.title = data['attributes']['title']
```
chrisseto: This should be taken care of by `BaseResource.__init__`

osfoffline/client/osf.py
```diff
@@ -101,19 +144,28 @@ def get_storage(self, id='osfstorage'):
...
 
+    def get_children(self, lazy=False):
+        related = map(
+            lambda data: Node(self.request_session, data),
+            self.fetch_related('children', {'embed': 'parent'})
+        )
+        if lazy:
+            return related
+        else:
+            return [item for item in related]
```
chrisseto: Try `list(related)`, it may be a bit more explicit.

osfoffline/client/osf.py
```diff
@@ -140,28 +192,34 @@ class Folder(StorageObject):
...
-            while data['links'].get('next'):
-                resp = self.request_session.get(data['links']['next'], params={'page[size]': 250})
-                data = resp.json()
-                l.extend(data['data'])
-            return [
-                (Folder if item['attributes']['kind'] == 'folder' else File)(self.request_session, item, parent=self)
-                for item in l
-            ]
-        return StorageObject(self.request_session, data['data'], parent=self)
+        self.name = data['attributes']['name']
```
chrisseto: Should be handled by `BaseResource.__init__`

osfoffline/client/osf.py
```diff
@@ -140,28 +192,34 @@ class Folder(StorageObject):
...
+        self.path = data['attributes']['path']
 
+    def __repr__(self):
+        return '<Folder {0} name={1} path={2}>'.format(id(self), self.name, self.path)
```
chrisseto: You may want to swap out Folder here with `__class__.__name__`

osfoffline/sync/ext/auditor.py
```diff
@@ -94,25 +120,65 @@ def context_for(paths):
...
+            )
+            for entry in Session().query(File)
+        }
 
     def collect_all_remote(self):
         ret = {}
         with ThreadPoolExecutor(max_workers=5) as tpe:
-            for node in Session().query(Node):
+            # first get top level nodes selected in settings
+            for node in Session().query(Node).filter(Node.sync == true()):
```
chrisseto: Any reason to not use `True`?
I think SQLAlchemy allows `.filter(Node.sync)`

osfoffline/sync/ext/auditor.py
```diff
@@ -130,20 +200,39 @@ def _collect_node_remote(self, root, acc, rel_path, tpe):
...
+                    child.extra['hashes']['sha256'],
+                    child
+                )
         tpe._work_queue.task_done()
         #return acc
 
     def collect_all_local(self, db_map):
         ret = {}
-        for node in Session().query(Node):
+        for node in Session().query(Node).filter(Node.sync == true()):
```
chrisseto: See above comment about `true()`

osfoffline/sync/remote.py
```diff
@@ -55,10 +67,12 @@ def run(self):
                 logger.info('Sleep interrupted, syncing now')
             self._sync_now_event.clear()
 
-            # Ensure selected node directories exist
-            for node in Session().query(Node).all():
-                local = Path(os.path.join(node.path, settings.OSF_STORAGE_FOLDER))
-                os.makedirs(str(local), exist_ok=True)
+            # Ensure selected node directories exist and db entries created
+            selected_nodes = list(
```
chrisseto: No reason to cast to a list here

start.py
```diff
@@ -9,6 +10,7 @@
 from osfoffline.gui.qt import OSFOfflineQT
 from osfoffline.utils.singleton import SingleInstance
 
+signal.signal(signal.SIGINT, signal.SIG_DFL)
```
chrisseto: IIRC this is a "dirty" shutdown.
You may want to disable this when not in debug mode.

osfoffline/client/osf.py
samchrisinger: I'm not sure that's the case for title.

samchrisinger: I see now

osfoffline/database/models.py
```diff
@@ -81,6 +105,14 @@ def validate_top_level(self, key, top_level):
             assert self.parent is not None
         return top_level
 
+    @validates('sync')
+    def validate_sync(self, key, sync):
+        if not self.parent_id:
+            assert sync is True
```
abought: Is there a realistic use case where these assertions might fire? We should avoid relying on them for production use- a custom exception class might be more helpful and more robust.

From https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement :
> Assert statements are a convenient way to insert debugging assertions into a program... The current code generator emits no code for an assert statement when optimization is requested at compile time.

samchrisinger: I chose to follow the existing convention used in other validators.

This validator is in place largely to improve feedback to developers who might make future assumptions about the meaning of Node.sync.

abought: As long as it's only internal and for debugging we can work with that; thanks.

The reason I brought this up at all is that we've been trying to move away from the existing convention in some legacy code, actually. :) It's an ongoing process...

PR #120, opened 2015-12-29
Fix windows sync issues

abought:
Refs [OFF-62](https://openscience.atlassian.net/browse/OFF-62)

## Purpose
Address problems with sync functionality on Windows operating systems.

## Summary of changes
- Replace several hardcoded OS-specific path separators with `os.path.sep`.
  - Fixes issue where first sync attempt lead to `parent is None` errors, and subsequent attempts lead to database integrity errors
- Add docstrings to clarify differences between methods.

## To do
- [ ] While testing, uncovered a secondary issue: on Windows, creating a new file in a folder fired an `Update` event, instead of a `Create` event. Although sync would complete correctly the second time through, this is not ideal. Problem did not occur on Mac OS and appears to be watchdog related.

- [x] Determine why a database state/uniqueness constraint error was issued. In theory, the database state should not change if the sync operation fails: the DB should not have had an entry for an operation that failed midway through. Seems symptomatic of a deeper issue.
    - This appears to be tied to the conflict resolution mechanism: it can be created with a forward slash in the path, but the auditor tries to find a DB object based on file path (not an exact string match), then creates a File object with a specific OSF id. (the file object was actually there, but not found by the auditor)

### DIFF
--- a/osfoffline/database/models.py
+++ b/osfoffline/database/models.py
@@ -49,7 +49,10 @@ class Node(Base):
 
     @property
     def path(self):
-        """Recursively walk up the path of the node. Top level node joins with the osf folder path of the user
+        """
+        Path on the local filesystem.
+
+        Recursively walk up the path of the node. Top level node joins with the osf folder path of the user
         """
         # +os.path.sep+ instead of os.path.join: http://stackoverflow.com/a/14504695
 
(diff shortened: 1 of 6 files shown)


abought:
Proposed plan: will deal with platform-specific watchdog differences in a separate PR. 

To keep QA moving, will advance this PR (fix for basic syncing operations) separately.

PR #134, opened 2016-01-19
[RFR] Handle remote deletion of resources

samchrisinger:
# Followup (for a later PR)

We should find a way to shut down the RemoteSyncWorker and OperationWorker without waiting for their queues to be flushed. When a user logs out we're dropping the database anyways, and letting the sync operations finish adds unnecessary work and delay to the logout process. 

# Purpose

see: 
https://openscience.atlassian.net/browse/OFF-85
https://openscience.atlassian.net/browse/OFF-86

tldr: deleting a project that was synced locally would cause the remote sync loop to hang

# Changes

- add ClientLoadError as a generic exception for API requests that fail for some reason
- try/except Node loads in RemoteSyncWorker's#_preprocess_node, and explicitly hander 4XX status codes
- some other cleanup from https://github.com/CenterForOpenScience/OSF-Offline/pull/125, in particular with consolidation/unification of how we shut down workers

### DIFF
--- a/osfoffline/client/osf.py
+++ b/osfoffline/client/osf.py
@@ -67,6 +67,12 @@ class BaseResource(abc.ABC):
     @classmethod
     def load(cls, request_session, *args, **kwargs):
         resp = request_session.get(cls.get_url(*args, **kwargs), params={'page[size]': 250})
+        if (resp.status_code % 500) <= 100:
+            raise ClientLoadError(
+                resource=cls.RESOURCE,
+                status=resp.status_code,
+                errors=['Server error']
+            )
         data = resp.json()
 
         if 'errors' in data:
--- a/osfoffline/client/osf.py
+++ b/osfoffline/client/osf.py
@@ -67,7 +67,7 @@ class BaseResource(abc.ABC):
     @classmethod
     def load(cls, request_session, *args, **kwargs):
         resp = request_session.get(cls.get_url(*args, **kwargs), params={'page[size]': 250})
-        if (resp.status_code % 500) <= 100:
+        if resp.status_code >= 500:
             raise ClientLoadError(
                 resource=cls.RESOURCE,
                 status=resp.status_code,
--- a/osfoffline/client/osf.py
+++ b/osfoffline/client/osf.py
@@ -11,6 +11,14 @@ from osfoffline.utils import Singleton
 from osfoffline.utils.authentication import get_current_user
 
 
+class ClientLoadError(Exception):
+
+    def __init__(self, *args, resource=None, status=None, errors=None):
+        super(ClientLoadError, self).__init__(*args)
+        self.resource = resource
+        self.status = status
+        self.errors = errors
+
 class OSFClient(metaclass=Singleton):
 
     def __init__(self, *, limit=5):
@@ -60,6 +68,13 @@ class BaseResource(abc.ABC):
     def load(cls, request_session, *args, **kwargs):
         resp = request_session.get(cls.get_url(*args, **kwargs), params={'page[size]': 250})
         data = resp.json()
+
+        if 'errors' in data:
+            raise ClientLoadError(
+                resource=cls.RESOURCE,
+                status=resp.status_code,
+                errors=data['errors']
+            )
 
         if isinstance(data['data'], list):
             l = data['data']
@@ -182,6 +197,13 @@ class StorageObject(BaseResource):
     def load(cls, request_session, *args):
         resp = request_session.get(cls.get_url(*args), params={'page[size]': 250})
         data = resp.json()
+
+        if 'errors' in data:
+            raise ClientLoadError(
+                resource=cls.RESOURCE,
+                status=resp.status_code,
+                errors=data['errors']
+            )
 
         if isinstance(data['data'], list):
             return [
--- a/osfoffline/sync/local.py
+++ b/osfoffline/sync/local.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import threading
 
 from watchdog.observers import Observer
+from sqlalchemy.orm.exc import NoResultFound
 
 from osfoffline import utils
 from osfoffline.utils.authentication import get_current_user
@@ -21,11 +22,19 @@ class LocalSyncWorker(ConsolidatedEventHandler, metaclass=Singleton):
 
     def __init__(self):
         super().__init__()
-        self.folder = get_current_user().folder
+        try:
+            user = get_current_user()
+        except NoResultFound:
+            user = None
+            self.folder = None
+
+        if user:
+            self.folder = user.folder
 
         self.observer = Observer()
         self.ignore = threading.Event()
-        self.observer.schedule(self, self.folder, recursive=True)
+        if self.folder:
+            self.observer.schedule(self, self.folder, recursive=True)
 
     def start(self):
         logger.debug('Starting watchdog observer')
--- a/osfoffline/sync/local.py
+++ b/osfoffline/sync/local.py
@@ -35,7 +35,7 @@ class LocalSyncWorker(ConsolidatedEventHandler, metaclass=Singleton):
         logger.debug('Stopping LocalSyncWorker')
         # observer is actually a separate child thread and must be join()ed
         self.observer.stop()
-        del type(self.__class__)._instances[self.__class__]
+        self.join()
 
     def join(self):
         self.observer.join()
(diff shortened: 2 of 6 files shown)


osfoffline/client/osf.py
```diff
@@ -59,8 +67,21 @@ def from_data(cls, request_session, data):
     @classmethod
     def load(cls, request_session, *args, **kwargs):
         resp = request_session.get(cls.get_url(*args, **kwargs), params={'page[size]': 250})
+        if (resp.status_code % 500) <= 100:
```
abought: For clarity, can you please use `if resp.status_code >= 500`?

(unless there is a benefit to using modulo here...?)

samchrisinger: catching 500, 502

abought: Right, but that would be covered by the inequality:

`(502 >= 500 and 500>=500) == True`
`resp.status_code >= 500`

I'm not seeing the functional/readability advantage of the modulo, I guess. (though I'll admit that it's one of my favorite fun operators)

osfoffline/sync/local.py
```diff
@@ -21,11 +22,19 @@ class LocalSyncWorker(ConsolidatedEventHandler, metaclass=Singleton):
...
+            user = None
+            self.folder = None
+
+        if user:
+            self.folder = user.folder
 
         self.observer = Observer()
         self.ignore = threading.Event()
-        self.observer.schedule(self, self.folder, recursive=True)
+        if self.folder:
```
abought: Can you help me understand this a bit better? It appears that if there is no user, it will never schedule an observer.

But:
1. Under what conditions does a `LocalSyncWorker` get scheduled if there is no logged-in user? It seems like this code shouldn't even be reachable if line 27 throws an exception.
2. I don't see a mechanism here for the observer to get scheduled once a user does become available. (and since this is a singleton, nothing short of logging out would lead to the creation of a new instance)

Does this line create a situation where watchdog simply wouldn't run for the user at all?

osfoffline/sync/remote.py
```diff
@@ -88,7 +89,11 @@ def run(self):
                 Notification().error(msg)
                 logger.exception(msg)
 
-            time.sleep(10)  # FIXME Per icereval, this is "due to watchdog not clearing its event evaluation when the lock is cleared"
+            # We need to ignore modifications to the local filesystem made by the RemoteSyncWorker.
+            # Since there can be a delay between when an operation is popped off the OperationWorker's
+            # queue and when the event is actually captured by watchdog, this sleep tries to ensure the
+            # watchdog observer does not capture any events triggered by the application itself.
```
abought: This is a much more informative source code comment. Thanks!

abought:
Using this branch, I still see the behavior described in [OFF-86](https://openscience.atlassian.net/browse/OFF-86)- the program hangs if I try to log out (from system tray menu) while a project is syncing.

Occurs on Mac OS, new DB/ first sync of single project. (Python 3.5, QT 5.5.1)

I waited for 1-2 min but the program remained frozen. (normally the project syncs in < 20 sec). Reproduced 2x.

abought:
Confirmed that the test case for OFF-85 crashes on develop, but appears fixed in this PR. Whee!

abought:
After testing 85, I did get a really odd error message when closing the application:

```
Exception in thread Thread-3:
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/threading.py", line 923, in _bootstrap_inner
    self.run()
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 76, in run
    self._preprocess_node(node, delete=False)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 134, in _preprocess_node
    self._orphan_children(node, stack)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 115, in _orphan_children
    for record in node.children:
  File "/Users/andyboughton/code/cos/OSF-Offline/env/lib/python3.5/site-packages/sqlalchemy/orm/attributes.py", line 237, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/Users/andyboughton/code/cos/OSF-Offline/env/lib/python3.5/site-packages/sqlalchemy/orm/attributes.py", line 583, in get
    value = self.callable_(state, passive)
  File "/Users/andyboughton/code/cos/OSF-Offline/env/lib/python3.5/site-packages/sqlalchemy/orm/strategies.py", line 505, in _load_for_state
    (orm_util.state_str(state), self.key)
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Node at 0x1142eceb8> is not bound to a Session; lazy load operation of attribute 'children' cannot proceed
```

osfoffline/sync/remote.py
```diff
@@ -107,14 +111,25 @@ def _orphan_children(self, node, remote_children):
...
 
     def _preprocess_node(self, node, *, delete=True):
         nodes = [node]
-        remote_node = OSFClient().get_node(node.id)
+        try:
+            remote_node = OSFClient().get_node(node.id)
+        except ClientLoadError as err:
+            if err.status in (http.client.NOT_FOUND, http.client.GONE):
+                # cascade should automagically delete children
+                Session().delete(node)
```
abought: It appears that calling delete doesn't actually delete from the underlying DB. You may need to use `Session().commit` or `flush` after this to avoid error messages and various other odd behavior.

abought: This is a somewhat destructive action that we may want to track. 

Can you please add a logging message (`logger.info`) to indicate that the project was dropped from the local sync database?

samchrisinger: See L55 for the commit().

abought: Any thoughts on the error message I got when closing the application, then? (I updated the comment to include a few more lines of context from before I clicked "quit")

osfoffline/sync/local.py
samchrisinger: That's a good point. For the time being letting the exception go unhandled is the best approach here.

osfoffline/sync/remote.py
samchrisinger: Looks like I needed a commit elsewhere (the other place _preprocess_node is called). Thank you for the catch.

abought:
After logging out and waiting several seconds at the start screen, I got the following message. It appears that something that uses `LocalSyncWorker` has not been completely stopped.

However, it doesn't hang the login screen, and when I do log in, things seem to work (I can log in, select projects, and Watchdog events fire). Noting for posterity but I won't block the merge.

osfoffline/sync/remote.py
```diff
@@ -107,14 +113,27 @@ def _orphan_children(self, node, remote_children):
         locally for which the remote Node has been deleted are explicitly removed
         from OSFO's auditing and will be ignored.
         """
-        children_ids = map(lambda c: c.id, remote_children)
+        children_ids = [c.id for c in remote_children]
         for record in node.children:
             if record.id not in children_ids:
                 Session().delete(record)
+                logger.info("Deleted remotely deleted database Node<{}>".format(record.id))
```
abought: `record` is undefined inside this function; this logging message fails with an exception when run, and locks the progrsm. (as a shortcut, running `oinv flake` will catch this error)

```
[INFO][2016-01-21 10:03:39,062][Thread-3][osfoffline.sync.remote]: Beginning remote sync
[INFO][2016-01-21 10:03:39,450][Thread-3][requests.packages.urllib3.connectionpool]: Resetting dropped connection: test-api.osf.io
Exception in thread Thread-3:
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 128, in _preprocess_node
    remote_node = OSFClient().get_node(node.id)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/client/osf.py", line 34, in get_node
    return Node.load(self.request_session, id)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/client/osf.py", line 82, in load
    errors=data['errors']
osfoffline.client.osf.ClientLoadError

During handling of the above exception, another exception occurred:

  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/threading.py", line 923, in _bootstrap_inner
    self.run()
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 76, in run
    self._preprocess_node(node, delete=False)
  File "/Users/andyboughton/code/cos/OSF-Offline/osfoffline/sync/remote.py", line 133, in _preprocess_node
    logger.info("Deleted remotely deleted database Node<{}>".format(record.id))
NameError: name 'record' is not defined


```

PR #145, opened 2016-01-29
Add: Global waiting cursor when fetching projects

barbour-em:
This makes the waiting cursor visible over all widgets as opposed to the hourglass that was just over the treewidget
[OFF-25]

### DIFF
--- a/osfoffline/gui/qt/preferences.py
+++ b/osfoffline/gui/qt/preferences.py
@@ -24,7 +24,7 @@ from osfoffline.sync.remote import RemoteSyncWorker
 logger = logging.getLogger(__name__)
 
 def waiting_effects(function):
-    @functools.wraps(function)
+    @wraps(function)
     def wrap(*args, **kwargs):
         QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
         try:
--- a/osfoffline/gui/qt/preferences.py
+++ b/osfoffline/gui/qt/preferences.py
@@ -1,5 +1,6 @@
 import os
 import logging
+from functools import wraps
 
 from PyQt5 import QtCore
 from PyQt5.QtCore import QCoreApplication
@@ -9,10 +10,10 @@ from PyQt5.QtWidgets import QDialog
 from PyQt5.QtWidgets import QFileDialog
 from PyQt5.QtWidgets import QMessageBox
 from PyQt5.QtWidgets import QTreeWidgetItem
+from PyQt5.QtWidgets import QApplication
 from sqlalchemy.orm.exc import NoResultFound
 
 from osfoffline import language
-from osfoffline.utils import waiting_effects
 from osfoffline.client.osf import OSFClient
 from osfoffline.database import Session
 from osfoffline.database.models import User, Node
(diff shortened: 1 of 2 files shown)


osfoffline/gui/qt/preferences.py
```diff
@@ -148,13 +149,14 @@ def selector(self, selected_index):
 
             self._executor = QtCore.QThread()
             self.node_fetcher = NodeFetcher()
-            self.treeWidget.setCursor(QtCore.Qt.BusyCursor)
+            QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
```
abought: Out of curiosity, will the application cursor get "stuck" in a waiting mode if the node-fetcher operation fails? (eg L207)

barbour-em: I would think so yes. For this reason I tried to write a decorator for the waiting cursor:

```python
def waiting_effects(function):
        def wrap(*args, **kwargs):
            QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
            try:
                function(*args, **kwargs)
            except Exception as e:
                raise e
                logger.exception(e.args[0])
            finally:
                QApplication.restoreOverrideCursor()
        return wrap
```

For some reason, however, none of the cursor changes happened with the above. Any other suggestions on how to avoid this?

abought: Interesting. It may depend on what function you tried to decorate. (QT signals and slots are kind of their own world under the hood, for example)

Also, you may want to move the `logger.exception` call above the `raise e` statement so it actually fires. :)

barbour-em: Sure enough, just had to move it around. Added in c9e41c0

osfoffline/utils/__init__.py
```diff
@@ -1,6 +1,9 @@
 import hashlib
 import os
 
+from PyQt5 import QtCore
+from PyQt5.QtCore import QCoreApplication
+from PyQt5.QtWidgets import QApplication
```
abought: One other request: for unit testing purposes, we'd like to keep QT code out of the general utils module (see the CI server failure report for an example). 

Consider moving this decorator to somewhere more UI-centric, rather than mixing UI logic with application logic. You'll note we try to import QT in relatively few places throughout the codebase.

That should be it for drive-by comments from me; I now return you to your regular reviewers. :)

barbour-em: By more UI-centric do you mean inside the `preferences.py` or in something like a `osfoffline/utils.py`?

chennan47:
Whenever the event is triggered it generates a dialog in terminal.
screen shot 2016-02-01 at 3 08 13 pm

barbour-em:
@chennan47, I get this terminal dialog even when commenting out the `waiting_effects` decorator. It looks like what is happening is that since `self._executor = QtCore.QThread` is being instantiated inside `#selector` it gets garbage collected when that method is done. I have worked on this some but can't figure out how to stop this

chennan47:
The original ticket https://openscience.atlassian.net/browse/OFF-25 suggests giving user notification instead of make the cursor global pending. Now the cursor will change everytime you click the panel when doing fetch project however sometimes unnecessary since it might only take a sec to fetch the only one project user have. And since you don't have a way to fix the Qthread problem, why not go with the notification fix and issue another ticket with low priority for change cursor to be a improvement task.

barbour-em:
By 'user notification' does that mean adding something like 
```python
Notification.info('We are fetching your list of projects. This may take a few minutes.')
```

to the #fetch method?

chennan47:
@barbour-em yes, that's the notification.

barbour-em:
Closed in favor of #154

PR #146, opened 2016-02-01
Allow user to change location of OSF folder

barbour-em:
Before, the Change button on the general settings page didn't do anything. Now it allows updates to the field in the text box and saves.

### DIFF
--- a/osfoffline/gui/qt/preferences.py
+++ b/osfoffline/gui/qt/preferences.py
@@ -1,5 +1,6 @@
 import os
 import logging
+from distutils.dir_util import copy_tree
 
 from PyQt5 import QtCore
 from PyQt5.QtCore import QCoreApplication
@@ -43,8 +44,7 @@ class Preferences(QDialog, Ui_Settings):
         self._translate = QCoreApplication.translate
 
         self.changeFolderButton_2.clicked.connect(self.update_sync_nodes)
-        self.changeFolderButton.clicked.connect(self.change_folder_from_edit_box)
-        self.browseFolderButton.clicked.connect(self.set_containing_folder)
+        self.changeFolderButton.clicked.connect(self.update_containing_folder)
         self.pushButton.clicked.connect(self.sync_all)
         self.pushButton_2.clicked.connect(self.sync_none)
         self.tabWidget.currentChanged.connect(self.selector)
@@ -55,9 +55,8 @@ class Preferences(QDialog, Ui_Settings):
         self._executor = QtCore.QThread()
         self.node_fetcher = NodeFetcher()
 
-    def change_folder_from_edit_box(self):
-        new_folder = self.containingFolderTextEdit.toPlainText()
-        self.set_containing_folder(new_folder=new_folder, save_setting=True)
+    def update_containing_folder(self):
+        self.set_containing_folder(save_setting=True)
 
     def closeEvent(self, event):
         if set(self.selected_nodes) != set([node.id for tree_item, node in self.tree_items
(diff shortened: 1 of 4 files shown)


chennan47:
screen shot 2016-02-01 at 2 58 48 pm
After change the folder and click on change button and then hit the close of the setting panel. This dialog pops out which only supposed to pop up when you have unsaved project to sync changes. And also change button may need a rename or some other action to inform user it is taking effect.

After you change the folder location and choose to sync file again, it will think the project is empty and then remove all the files under the project on OSF.

barbour-em:
@chennan47, my solution for this was to sync the checked nodes after the folder location is changed. I also added a prompt to make sure the user knows what could happen. It is also more obvious what the button does.

chennan47:
@barbour-em This change folder button's naming is not ideal. Usually user will expect click that button to browser through the folder to select one to change instead of a 'confirm change folder' button as what it really does. Besides the UI is broken.
screen shot 2016-02-03 at 11 25 29 am


And still a more serious issue. If I choose to sync project A and afterwards, I choose to change folder and then sync project A again. As the folder path is changed however all the files renames in old folders. It thinks you have removed all the files in your local, as a result when you click 'sync now' it will remove all the files under project A from OSF. So actually when you change the folder path, you also need to copy all the selected syncs nodes files into the new folder.

chennan47:
Get it work once, then on Mac I got error message.
screen shot 2016-02-08 at 1 44 07 pm
And then nothing happen but return me back to the settings window.

On windows, Notification pops up saying error updating files which I just did a folder change which should not update the file.

The new UI interface looks neat and nice work, but since the text input route change for the folder won't do anything. It will be better to remove the input text field and let the change button take control. Also we need better notification to let user know they successfully update the folder or failed.

barbour-em:
I changed the text input to read only, this way the user can still see the full path to the folder. I also added notifications for success or fail of the folder copy.

As for the error, I haven't been able to make my OS X or Windows machine do this. What did you do to cause it?

osfoffline/gui/qt/preferences.py
```diff
@@ -84,12 +89,22 @@ def set_containing_folder(self):
             return
 
         user = Session().query(User).one()
+
+        if save_setting:
+            # copy the synced folders from the old to new location
+            # so OSF doesn't think they were deleted
+            copy_tree(user.folder, os.path.join(osf_path))
```
chennan47: If I put some projects which is not in the sync list or just some random folders inside the original path '/User/Desktop/OSF' for example. It will break as it can't recognize these unrelated folder which it should not copy and also if I choose the new destination to be '/User/Desktop/OSF/OSF' it will copy a new OSF folder into the current directory.

osfoffline/gui/qt/preferences.py
```diff
@@ -84,12 +89,22 @@ def set_containing_folder(self):
...
+
         user.folder = os.path.join(osf_path)
 
         self.containingFolderTextEdit.setText(self._translate("Preferences", self.containing_folder))
```
chennan47: Since the textedit field doesn't actually have a text edit usage. Change https://github.com/CenterForOpenScience/osf-sync/blob/develop/osfoffline/gui/qt/generated/preferences.py#L38 to user Qlabel instead of QTextedit and change the naming of the related field to avoid confusion.

abought: Small aside: anything in the generated folder should probably be changed in QT Designer; otherwise the python-code changes might get lost next time someone made a change using QT Designer.

chennan47:
close this in favor of https://github.com/CenterForOpenScience/osf-sync/pull/159

PR #170, opened 2016-02-18
[feature] db session contextmanager

icereval:


icereval:
Needs functional review, basic testing shows a few initial issues.

chennan47:
After cleanup db and login and then choose a project to sync, get this error and after a while app quit itself.
screen shot 2016-02-18 at 9 48 01 am

chennan47:
I still got the error message in the log above, but now it doesn't quit the app. However, sync projects is slow and sometimes freeze.

osfoffline/database/__init__.py
```diff
@@ -30,4 +32,5 @@ def drop_db():
 
 def clear_models():
     for model in CORE_OSFO_MODELS:
-        Session().query(model).delete()
+        with Session() as session:
```
chrisseto: The for loop should be within the context manager

PR #187, opened 2016-05-17
Feature/tmpfiles

chrisseto:


osfoffline/sync/ext/watchdog.py
```diff
@@ -1,242 +1,83 @@
...
+    if event.is_directory:
+        return None
+
     try:
         node = utils.extract_node(event.src_path)
-    except NodeNotFound:
-        db_file = None
-    else:
         db_file = utils.local_to_db(event.src_path, node, check_is_folder=False)
+        if db_file:
```
samchrisinger: Can you `try` to keep this out of the try/except block?

osfoffline/sync/ext/watchdog.py
```diff
@@ -1,242 +1,83 @@
...
-                        #   eg saving in vim or replacing a file in finder.
-                        Event = DirModifiedEvent if event.is_directory else FileModifiedEvent
-                        event = Event(event.src_path)
-                self._event_cache[parts] = event
-
-            logger.debug('Create cache: {}'.format(self._create_cache))
-            logger.debug('Event cache: {}'.format(self._event_cache))
+            try:
+                event.sha256 = sha256_from_event(event)
+            except FileNotFoundError:
```
samchrisinger: Might be good to move this to `sha_256_from_event` (other error handling is already happening there)

osfoffline/sync/local.py
```diff
@@ -107,28 +110,27 @@ def on_created(self, event):
...
     def on_deleted(self, event, *args, is_folder=False, **kwargs):
-        logger.info('Deletion event for {}: {}'.format('directory' if event.is_directory else 'file',
-                                                       event.src_path))
+        logger.info('Deletion event for {}: {}'.format('directory' if event.is_directory else 'file', event.src_path))
         # A hack: override checking if the passed path is a directory. Since Windows
         # emits folder deletion events as file deletes we need to ignore whether or not
         # a delete event is for a folder. Since the RemoteDelete operation works identically
         # for files and folders we can get away with this here.
         context = OperationContext(local=Path(event.src_path), check_is_folder=False)
+        context.db  # noqa
```
samchrisinger: ew

osfoffline/sync/utils.py
```diff
@@ -0,0 +1,154 @@
...
+        self._hash_pool.clear()
+
+    def push(self, event):
+        self._events.append(event)
+
+        self._push(event.src_path, event)
+        if event.event_type == EVENT_TYPE_MOVED:
+            self._push(event.dest_path, event, self._pool[event.src_path])
+
+    def _push(self, path, event, item=None):
```
samchrisinger: It might be clearer to handle the case where `item` is explicitly passed first.

samchrisinger:
I think we're dealing with the harsh reality that there will be at least one terrible function to handle event consolidation. My comments are all related to readabilty, and I won't claimed to have grokked the branchiing logic in `_push`, but overall this seems like an improvement.

chrisseto:
4 `if`s are better than 30 👍

samchrisinger:
wooo

PR #194, opened 2016-08-11
handle moving files to newly renamed folder

skurfer:
This mostly fixes SYNC-135. See the comment for an explanation of the fix.

Where it breaks down is that we can’t find the correct “delete” event 100% of the time. Specifically if there’s more than one “delete” event present, and the filename associated with the event we need has also been changed.

We were also hoping to fix SYNC-120 here, and were pretty close, but it’s more complicated than we thought. Let me know if you’d like to see where we were heading with that. (We thought we had it fixed, but there seems to be a race condition somewhere. Pausing to look at things in the debugger for a while will make the tests pass, but actually running them doesn’t work.)

### DIFF
--- a/osfoffline/sync/utils.py
+++ b/osfoffline/sync/utils.py
@@ -182,16 +182,33 @@ class EventConsolidator:
 
         item = self._pool.setdefault(path, item or Item(event.is_directory))
 
-        if copy_found and event.event_type == events.EVENT_TYPE_DELETED and item.events[-1].event_type == EVENT_TYPE_MOVED and path not in {item.events[-1].src_path, item.events[-1].dest_path}:
-            # Handles the lovely case of:
-            #   Create a folder
-            #   Rename folder
-            #   Move a file into that folder
-            # Due to limitations in inotify and the windows API, this chain of events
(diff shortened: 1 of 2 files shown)


skurfer:
A rather important detail I should have included: This will only work with a build of Watchdog that identifies synthesized events.

You can see the small change we made on our fork of the project.

https://github.com/sixfeetup/watchdog/commit/0a4b21ebaf126e418e508c839cd80d1d72b259c9

chrisseto:
Writing a note-to-self here: Update `requirements.txt` with sixfeetup's fork of watchdog

skurfer:
We could (probably should) do a pull request to get it into the official package, but they don’t seem to be merging and releasing very quickly, so you’ll probably still have to do this. I’ll bring it up in our daily meeting.

chrisseto:
@skurfer I'm going to pin to https://github.com/sixfeetup/watchdog/tree/synthetic for the moment, if that is okay with you.

skurfer:
Pin away. We can keep that branch out there indefinitely. I sent them a pull request, so hopefully it’ll end up in the official release.

